Issue #3, opened 2020-12-21
add readme


vykster:
Readme should include
- [x] sufficient information to deploy locally by another developer
- [ ] names of contributors, esp those who are not on GitHub

PR #24, opened 2021-07-13
Config for running site locally

### LINKED ISSUE #23: Configure running on localhost
We should be able to run this site locally in a known environment, regardless of operating system and local resources. This will help in testing updates to dependencies, bugfixes, and other feature development.

A simple Docker Compose set up would be sufficient to start.

thekaveman:
Closes #23.

Partially addresses #3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To run the site locally with `docker-compose`:
 docker-compose up [-d] site
 ```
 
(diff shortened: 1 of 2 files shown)


allejo (approved):

thekaveman (commented):
Funny, I copy/pasted this from the mobi-mart README. So the funky grammatical error is over there too 🙄

PR #109, opened 2023-05-15
chore: add .DS_Store ignore for Macs

machikoyasuda:
This is a chore PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 .bundle
 .netlify
 _site
-**/.DS_Store
+.DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 .bundle
 .netlify
 _site
+**/.DS_Store
```
thekaveman, line 5:
```suggestion
.DS_Store
```

This is a file right? Just the filename should catch it at any level of the repo.

thekaveman (approved):

PR #111, opened 2023-05-15
Refactor: Footer

### LINKED ISSUE #99: Implement new design and links for Footer
The footer has a new link ("Back to top"), and some new CSS styles

machikoyasuda:
closes #99 

## What this PR does
- Move footer to its own file
- Add 1 new link to footer
- Change footer link font size
- CSS grid fixing for mobile

## Screenshots

image

image

image

## Figma

For mobile, the designs call for 3 links in the first column, and 1 on the second...... but I cannot figure out how to do that with CSS Grid

image


machikoyasuda:
This is now ready for review @angela-tran @thekaveman

src/_layouts/default.html
```diff
@@ -7,4 +7,6 @@
   <body>
     {% include header.html %} {{content}}
   </body>
+
+  {% include footer.html %}
```
thekaveman, line 11: Nice idea to make this an include! I think it should be inside the `<body>` tag though, right?

thekaveman (changes requested):

src/_layouts/default.html
thekaveman: Here's a good explanation of why: https://stackoverflow.com/a/12107279/453168

thekaveman (approved):
LGTM

angela-tran (approved):

PR #122, opened 2023-05-17
Refactor: Navbar header

machikoyasuda:
closes #98 

## What this PR does
- [x] Remove old header, old header CSS
- [x] Styling for desktop 
- [x] Styling for mobile
- [x] Adds bootstrap.js dependency to main layout
- [x] Creates new nav with Bootstrap Navbar
- [x] Adds 2 placeholder links
- [x] New better logo
- [x] Fixes issue where the main container goes under the header
- Introduces `container` widths to the app. The rest of the app uses `80vw` to set its width, so we will have to slowly take all that CSS out and use `container`/`col`/`row` Bootstrap logic

## Screenshots

image

image

image
image

## Caveat

image
This is not the same as the Figma boards 
image
But -- this will put Calitp.org at parity with MobiMart. MobiMart doesn't have the custom hamburger menu icon, either. It can be swapped out on both apps in future PRs.

PR #118, opened 2023-05-16
Feat: define Jekyll collection for Press

thekaveman:
Closes #106 

Note I moved the styling of the Press Release list on the landing page to #107 and split out the styling of the Press Release page to #140.

PR #133, opened 2023-05-17
Refactor: Use Bootstrap container, remove article & constricted

### LINKED ISSUE #123: Refactor: Remove article padding / use Bootstrap cols

machikoyasuda:
closes #123 

### Removes Article
- Removes use of `<article>`. 
- Deletes `article` CSS.
- Uses `<main class="container">` in `default.html` layout instead. All pages should now use `<section class="row">` and `<div class="col-6">` syntax logic.

### Removes Constricted
- Removes use of `class="constricted"` and replaces it with `row/col/justify-center`
- Deletes `.constricted` CSS.

## What this PR does NOT do
- Refactor out CSS Grid from Footer, Header (to come in #122), and any other homepage content sections (beyond the "constricted" style)

### Removes Article refactor
image
Note: This header top issue will be fixed in #122 

### Removes Constricted refactor
image
image
<img width="1512" alt="image" src="https://github.com/cal-itp/calitp.org/assets/3673236/135276e3-4ded-4a70-9957-7f942767a131"

### DIFF
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -5,6 +5,8 @@
   </head>
 
   <body>
-    {% include header.html %} {{content}} {% include footer.html %}
+    {% include header.html %}
+    <main class="container">{{content}}</main>
+    {% include footer.html %}
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -3,266 +3,259 @@ title: "Example"
 layout: default
 ---
 
-<article>
-  <section id="deck">
-    <section>
-      <h1>A modern and consistent transportation experience throughout California</h1>
-      <p>
-        Learn how the California Integrated Travel Project (Cal-ITP) is making riding by bus and train simpler and more
-        cost-effective—for providers and customers.
-      </p>
-    </section>
-
-    <picture>
-      <img
-        id="triforce"
-        src="images/hero-header.png"
-        alt="A trio of images, clockwise from top: a bus, a train platform with a sign that announces “Next train in 3 minutes,” and a transit rider paying their fare by tapping their smartphone’s mobile wallet on a payment reader when boarding"
-      />
-    </picture>
+<section id="deck">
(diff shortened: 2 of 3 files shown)


machikoyasuda:
@thekaveman After this PR is merged, rebase this PR with the latest from `staging` https://github.com/cal-itp/calitp.org/pull/118/files and the contents of the Press page should be in a `container`. The Header will still overlap with it until #122 is merged.

thekaveman (changes requested):
This looks mostly really good!

I think using `<section>` for everything here is not semantically correct though. From the [spec](https://html.spec.whatwg.org/multipage/sections.html#the-section-element):

> The [section](https://html.spec.whatwg.org/multipage/sections.html#the-section-element) element is not a generic container element. When an element is needed only for styling purposes or as a convenience for scripting, authors are encouraged to use the [div](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element) element instead. A general rule is that the [section](https://html.spec.whatwg.org/multipage/sections.html#the-section-element) element is appropriate only if the element's contents would be listed explicitly in the document's [outline](https://html.spec.whatwg.org/multipage/sections.html#outline).

The highest-level elements as `<section>` e.g. `<section id="deck">`, `<section id="details">` seem appropriate, but within those I think `<div>` might be the better choice.

Thoughts?

machikoyasuda:
@thekaveman Yeah - having `<section>`s _inside_ other `<section>`s doesn't make any sense. I think we should change the children `<section>`s in subsequent PRs that actually change the contents. This PR is only concerned with the outermost `<main>` parent (grandparent?).

cc @angela-tran

angela-tran:
I agree we should change the inner `section`s to `div`s. I'm ok with that being done as a separate issue

angela-tran (approved):

thekaveman (approved):

Issue #90, opened 2023-05-04
Figure out implementation for filtering


angela-tran:
This task is to figure out a clean implementation for the "filtering pills" shown on designs for the new "Press" and "Resource" pages.

Some questions to answer are:
 - Do we want to do this with a CSS-only approach? Do we need Javascript?
 - How should the data (links to press releases and resources) be modeled to support the filtering implementation?

@segacy1 confirmed in Slack that the user should be able to select multiple pills.

The result of this task is not a pull request that gets merged in, but rather a clear pattern for implementers of the "Press" and "Resource" pages to follow.

![image](https://user-images.githubusercontent.com/25497886/236339463-392d1058-bc37-4baf-9496-b42fc7ded77b.png)

angela-tran:
Tagging #83 as related

thekaveman:
> [Navs and tabs · Bootstrap v5.3](https://getbootstrap.com/docs/5.3/components/navs-tabs/#pills-with-dropdowns) + https://getbootstrap.com/docs/5.3/components/navs-tabs/#javascript-behavior  like the Contact Us page on mobimart.

thekaveman:
[UX conversation on Slack](https://cal-itp.slack.com/archives/C018XMWT2FN/p1684180778079749)

thekaveman:
## UX Summary

- Each pill is either `on` or `off`
- Only one pill can be `on` at one time
- Turning one pill `on` turns all others `off`
- Turning one pill `on` filters the list to items matching that pill
- Turning a pill `off` removes its filter
- The pill background is filled for a pill that is `on`
- The pill background is transparent for a pill that is `off`

angela-tran:
This is what I had gotten working before the Compiler retreat week:

![calitp-filtering](https://github.com/cal-itp/calitp.org/assets/25497886/4dfaa3fe-aed7-43b2-89c5-f5995cb47531)

This uses some of the WIP code from https://github.com/cal-itp/calitp.org/pull/142 to style the pills. I wrote some Javascript to implement turning a pill `off` since that is not provided by Bootstrap. I'll push the WIP code soon.

I think this meets all the items listed in the comment above under "Summary".

angela-tran:
In the Slack convo about the UX, we said that when the page loads, all pills are off. What I gather from the Slack thread is that the user sees an unfiltered list of all articles when all pills are off. I have some idea of the Javascript needed to accomplish this and will work on it today

angela-tran:
[...]

PR #148, opened 2023-05-31
Filtering spike

angela-tran:
Opening this PR mainly just so others can play with the filtering on the preview site.

### Notes
This is a very rough proof-of-concept for how we could accomplish the filtering UX described in #90.

There is custom Javascript to handle the parts that Bootstrap doesn't provide. Those parts are
 - turning a filter off
 - showing/hiding a tab that contains all articles

I think the JS code can be cleaned up / simplified. Also, there are definitely parts in `press.html` that would make sense to extract out as includes.

#### Handling no Javascript situation
[Bootstrap's notes on disabled Javascript](https://getbootstrap.com/docs/5.3/getting-started/javascript/#disabled-javascript)

This PR roughly shows how we could hide the pills when Javascript is not available. The tab with all articles is what shows, so the user experience is appropriate I think.

### DIFF
--- a/src/_includes/press_listing.html
+++ b/src/_includes/press_listing.html
@@ -1,0 +1,17 @@
+  {% comment %} Jekyll sorts by the date field in ascending order by default {% endcomment %}
+  {% assign all_press = site.press | reverse %}
+  {% for press in all_press %}
+    {% assign url = press.url %}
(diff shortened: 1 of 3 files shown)


angela-tran:
Meant to open this as a draft. Not intending this to be merged in

PR #151, opened 2023-06-01
Resources, Press pages

### LINKED ISSUE #105: Resources landing page
- [ ] Blue box + heading + intro content
- [ ] Pills
- [ ] Styles for list of Resources

### LINKED ISSUE #107: Press landing page
- [ ] Blue box + heading + intro content
- [ ] Pills
- [ ] Styles for list of Press Releases

machikoyasuda:
closes #107 
closes #105 

## What this PR does and doesn't do
- This PR does not have the Filter bubbles, but it has everything else.
- Rhombus! Rhombus is now only on window size over 992px.
- Creates 2 includes for the style and logic of showing items, showing `Show more` after 6 items
- Adds a few CSS utility classes, variables

### DIFF
--- a/src/_includes/articles.html
+++ b/src/_includes/articles.html
@@ -1,11 +1,11 @@
 {% for item in items limit: 6 %}
-  {% include article.html %}
+  {% include article.html item=item %}
 {% endfor %}
 {% if group.size > 6 %}
   <details>
     <summary class="d-block mb-3 pb-4">Show more</summary>
     {% for item in items offset: 6 %}
-      {% include article.html %}
+      {% include article.html item=item %}
     {% endfor %}
   </details>
 {% endif %}
--- a/src/_includes/articles.html
+++ b/src/_includes/articles.html
@@ -3,7 +3,7 @@
 {% endfor %}
 {% if group.size > 6 %}
   <details>
-    <summary>Show more</summary>
+    <summary class="d-block mb-3 pb-4">Show more</summary>
     {% for item in items offset: 6 %}
       {% include article.html %}
     {% endfor %}
--- a/src/_includes/articles.html
+++ b/src/_includes/articles.html
@@ -1,0 +1,11 @@
+{% for item in items limit: 6 %}
+  {% include article.html %}
+{% endfor %}
+{% if group.size > 6 %}
+  <details>
+    <summary>Show more</summary>
+    {% for item in items offset: 6 %}
+      {% include article.html %}
+    {% endfor %}
+  </details>
+{% endif %}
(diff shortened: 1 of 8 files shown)


machikoyasuda:
@segacy1 Does this `Show more` implementation look good? https://deploy-preview-151--cal-itp-website.netlify.app/resources I don't want the `Show more` to hide after it's open, cos then users won't be able to close it.

src/_includes/articles.html
```diff
@@ -0,0 +1,11 @@
+{% for item in items limit: 6 %}
+  {% include article.html %}
+{% endfor %}
+{% if group.size > 6 %}
+  <details>
+    <summary class="d-block mb-3 pb-4">Show more</summary>
+    {% for item in items offset: 6 %}
+      {% include article.html %}
```
thekaveman, line 8: Same comment about explicitly passing `item` to the include.

src/resources.html
```diff
@@ -3,31 +3,40 @@
...
-      </article>
-    {% endfor %}
-    {% unless forloop.last %}
-      <hr/>
-    {% endunless %}
-  {% endfor %}
+      {% for group in groups %}
+        <h2 class="mb-4 mt-5">{{ group.name }}</h2>
+        {% assign items = group.items %}
+        {% include articles.html %}
```
thekaveman, line 33: Same comment about explicitly passing to include.

thekaveman (changes requested):

segacy1:
@machikoyasuda what do you think about [this design](https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?type=design&node-id=6333-35170&t=djHz5aPCnjITVUUM-4) (show more/show less)? From what I understand, we don't have much content, so I don't think it would get unwieldy.

src/_includes/articles.html
thekaveman: @machikoyasuda I don't think this one was updated yet.

src/resources.html
thekaveman: Yep this looks good!

thekaveman (approved):
Code and staging site look good to me.

I feel like we shouldn't block this on the "Show more" UX. For this sprint/iteration we probably don't need it as the content isn't overwhelming for now (and sounds like we are even going to drop a Resource based on a comment from @mrose914).

Thoughts about delaying this UX for the next round?

machikoyasuda:
@thekaveman Sounds good to me.

thekaveman:
Errr... I meant we should remove it entirely. Not a big deal.

machikoyasuda:
@thekaveman Oops. Okay I can do that in a next PR!

PR #154, opened 2023-06-06
Remove Show More for now

machikoyasuda:
Will keep this `includes` file for adding back `Show More` in the future, after this sprint/iteration.


thekaveman (approved):
Thank you! 👍

Issue #92, opened 2023-05-04
Implement Press feature


angela-tran:
Depends on #90 for the filtering part

The styling and content for this page can be worked on independently from the filtering part.

Figma design: https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?type=design&node-id=5305-29411&t=qvrYiPrTQX0GD3Vm-4

## Tasks

- [x] #82 
- [x] #106
- [x] #140
- [x] #107
- [ ] Filters from #90

angela-tran:
Tagging https://github.com/cal-itp/calitp.org/issues/82 as related

Issue #91, opened 2023-05-04
Implement Resources feature


angela-tran:
Depends on #90 for the filtering part

The styling and content for this page can be worked on independently from the filtering part.

Figma design: https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?type=design&node-id=5305-30392&t=qvrYiPrTQX0GD3Vm-4

## Tasks

- [x] #104
- [x] #105
- [ ] Filters from #90
- [x] ~"Show More" link~ Leaving this until the next Sprint

angela-tran:
@mrose914 just a heads up that we will need the actual list of Resources (and their links) to put on this page

thekaveman:
Thank you @mrose914 those edits are helpful! 

One last question: the Figma designs show a date next to each resource, but what the date should be is not obvious for each of them. Can you please help getting a date assigned to each row in the sheet?

(This is more obvious for the existing press so you can ignore those).

thekaveman:
@mrose914 one last question for you here:

The Categories in the spreadsheet include:

- Case Study
- Fact Sheets
- Slide Decks
- One-Pager

Should these correspond exactly to the section headings on the landing page?

E.g. in the following screenshot the section headings are:

- Case Studies
- Fact Sheets & Overviews
- Testimonials (I know we aren't using this one right now)

![image](https://github.com/cal-itp/calitp.org/assets/1783439/dca1d9bd-2f2a-47f5-b55a-39bf77c1c6fe)

Based on the what is in the sheet, do these categories make sense (all pluralized nouns)?  Should any of these be grouped together (like Fact Sheets & Overviews in the screenshot)?

- Case Studies
- Fact Sheets
- Slide Decks
- One-Pagers

PR #163, opened 2023-06-08
Feat: article tabs

angela-tran:
Adds tabs to Press and Resource pages that contain articles based on tag.

~Note that this PR doesn't have logic to hide the tab with all articles, so it always appears.~ hiding the "all" tab for now

Part of #91 and #92

### DIFF
--- a/src/_includes/articles.html
+++ b/src/_includes/articles.html
@@ -1,3 +1,5 @@
 {% for item in items %}
-  {% include article.html item=item %}
+  {% if include.tag == empty or include.tag and press.tags contains include.tag %}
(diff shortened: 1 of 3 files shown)


machikoyasuda (approved):

thekaveman (commented):
Are we hardcoding the pills just to get this initial HTML structure in? Will we eventually use some `group_by` logic to be able to get the pills generated from the data?

machikoyasuda:
@thekaveman Yeah it can be cleaned up with that and `includes`

angela-tran:
@thekaveman Yes, that's correct

PR #164, opened 2023-06-08
Feat: toggling pills

angela-tran:
This PR implements logic to turn a pill off and back on again. When pills are off, then all articles should show.

Part of #91 and #92


angela-tran:
I was initially thinking of opening separate PRs for each piece of functionality (toggling pills, "all" tab behavior, and ensuring arrow keys work the same as mouse-clicks and Enters), but then I decided it makes more sense to have it all together

thekaveman (commented):
Feel free to disregard, but I noticed the formatting on `pill-behavior.html` was a little funny (some of the comments specifically) -- just wondering if it would be easier to manage this file as a `.js` script file vs. an include/HTML?

thekaveman (commented):
Looks like it works as intended! Keyboard and mouse navigation looks good!

We may want to check with Marissa/Segacy in a follow up if we should hide sections that don't have any items in Resources (e.g. for Benefits the Case Studies section is empty). Not something that needs to be addressed here though.

thekaveman (approved):

angela-tran:
Thanks @thekaveman ! I like your idea to manage the filtering logic as a `.js` file. Do these changes look ok: https://github.com/cal-itp/calitp.org/pull/164/commits/e94ce261b8de05db1c76ed9e5268371b144def90?

thekaveman (approved):
🎸

PR #165, opened 2023-06-08
Refactor: tag data

angela-tran:
Extract out tags to data files.

Part of #91 and #92


machikoyasuda (approved):
- tested both pages in keyboard mode and mouse mode

machikoyasuda:
Note for future QA and future improvements: This is the kind of page/interaction that would make having a Skip Nav super useful, to be able to skip the nav buttons while trying to change tabs/filters. Can be part of future improvements after this sprint though.

PR #167, opened 2023-06-08
Refactor: Use flexbox for footer, not css grid

machikoyasuda:
Part of CSS implementation/refactor

### Desktop
image

### Tablet
image

### Mobile
image

### DIFF
--- a/src/_includes/footer.html
+++ b/src/_includes/footer.html
@@ -1,33 +1,35 @@
-<footer>
-  <nav>
-    <ul class="links">
-      <li>
-        <a
-          class="fw-normal text-decoration-none text-white"
-          rel="noreferrer"
-          href="#">Back to top</a>
-      </li>
-      <li>
-        <a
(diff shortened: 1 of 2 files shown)


machikoyasuda:
@angela-tran @thekaveman I promise this is the last CSS refactor 😅 

Now the CSS code base is 100% free of CSS Grid (which is powerful, but not necessary for what this app required), there is no extra CSS required for the Footer (which also allowed us to delete a lot of media query code) - which makes this footer code also easier to copy to other apps. The `main.css` file now is pared down to only variables, variable overrides, new utility classes and reusable classes for _new_ styles created for this app (like rainbow lists, new nav pill styles, etc).

angela-tran (approved):
I compared this PR's preview site to what's currently on staging and see no regressions. The alignment looks better actually, which makes sense since it's all using Bootstrap now.

Code changes look good too

Issue #178, opened 2023-06-12
Updating spacing for homepage Contact rhombuses


thekaveman:
From https://github.com/cal-itp/calitp.org/issues/157#issuecomment-1585073506

- [ ] make top/bottom padding for text 24px #192
- [x] on Mobile, make rectangles full width to our normal content margin (i.e. within container) #188

PR #188, opened 2023-06-14
Fix: Fix Desktop rhombus bug on Safari

### LINKED ISSUE #174: Investigate split rectangle on Press/Resources in Mobile
@segacy1 notes that in Chrome on an iPhone 12 there is a funky split between the top paragraph and the pills in the colored rectangles:

split rectangle

@thekaveman was unable to reproduce in Chrome/Firefox responsive view at 320px and higher.

machikoyasuda:
closes #174 fully

- Used https://yoksel.github.io/relative-clip-path/ to create clip-paths directly from svg's from Figma. Tried using https://www.plantcss.com/css-clip-path-generator but the shapes were not coming out the same.
- Avoid use of `filter: url` entirely, b/c it's very buggy on Safari.
- Creates an includes for the Press and Resources SVG, as they use the same one.
- Adds SVGs in-line for all 3 Index page SVGs.
- Adds back box-shadow (using filter) on the bottom 2 rhombuses on Home page (https://css-tricks.com/using-box-shadows-and-clip-path-together/)
- This PR also fixes: the second point on #178

## How to test
- Apple/Mac OS: test on Safari, Firefox, Chrome for rhombus fixes
- Windows: test on Firefox, Chrome for no regressions
- Android: test on Chrome for no regressions
- iPhone: test on Safari for no regressions

## Before & After

image
image
Safari behaves very strangely with `filter: url()` -- the browser is trying to apply the filter to _everything_, from the link hover effect, to even the way the user highlights the text. There is no way to apply the filter to just one element in Safari, the browser will apply it to every child element and even highlighted text. And there's no way to add another filter to undo a url filter.

|Before   | After  |
|---|---|
|image|  image |  
|Totally bonkers.   |Normal.|  

image

image

### DIFF
--- a/src/_includes/clipped.html
+++ b/src/_includes/clipped.html
@@ -1,0 +1,3 @@
+<svg class="position-absolute">
+  <clipPath id="my-clip-path" clipPathUnits="objectBoundingBox"><path d="M0,0.081 C-0.001,0.051,0.007,0.026,0.018,0.026 L0.981,0 C0.992,0,1,0.025,1,0.055 L0.981,0.902 C0.981,0.928,0.973,0.949,0.963,0.95 L0.042,1 C0.033,1,0.025,0.98,0.024,0.953 L0,0.081"></path></clipPath>
+</svg>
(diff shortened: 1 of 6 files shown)


thekaveman:
@segacy1 when you have some time, can you please review this follow-up to the split rhombuses in Mac/Safari -- this time for Desktop widths. (And please ensure no regressions in Mobile as well).

thekaveman (changes requested):
Regressions on homepage Contact rhombuses, in Windows/Desktop Firefox and Chrome:

* Bottoms are no longer aligned
* The direction changed and no longer represents the Figma design

**Figma**

image

**Staging**

Bottom of each rhombus is aligned, direction "towards" each other as shown in Figma:

image

**This PR**

Bottom of the left rhombus is lower than the bottom of the right rhombus, direction "away" from each other:

image

machikoyasuda:
Fixes pushed:

- Switches the shapes around
- Uses flex-box to manage / keep the heights the same
- Adds back shadow

Safari:
image
image

Firefox:
image
image

segacy1 (approved):

thekaveman (approved):
Looks great! Confirmed the fix in Desktop Firefox/Chrome on Windows.

Also, thank you for removing that inline style definition 😁

Issue #177, opened 2023-06-12
Update spacing in Initiatives


thekaveman:
From https://github.com/cal-itp/calitp.org/issues/157#issuecomment-1585073506

- [x] center content within the rhombus in Desktop/Tablet #193
- [x] adjust vertical spacing for all sections to 40px Desktop, 28px Mobile #194
- [ ] ensure 8px of padding between all Recent News/Resources links #198

PR #194, opened 2023-06-14
Fix: initiatives sections spacing

angela-tran:
Part of https://github.com/cal-itp/calitp.org/issues/177

> adjust vertical spacing for all sections to 40px Desktop, 28px Mobile

### DIFF
--- a/src/_includes/initiatives.html
+++ b/src/_includes/initiatives.html
@@ -48,7 +48,7 @@
               alt="{{ initiative.image.alt }}"
               width="{{ initiative.image.width }}" />
           </picture>
-          <div class="col-12 col-md-8 col-lg-8 initiatives-section-mb">
+          <div class="col-12 col-md-8 col-lg-8 mb-28 mb-md-40">
             <h3 class="h2 text-white text-center text-md-start pb-4">{{ initiative.headline }}</h3>
             {% for paragraph in initiative.paragraphs %}
               <p class="text-white">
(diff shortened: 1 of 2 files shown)


angela-tran:
Going to rebase

thekaveman (commented):
I'm OK with introducing this helper class since it seems like 40px is not exactly within the Bootstrap `mb-*` framework. Wondering if @machikoyasuda has any thoughts on this approach?

Also, nitpick, but it feels a little strange to me to have the margin applied to the Initiatives button vs. its parent `div`, but this also works, so I won't block because of it.

machikoyasuda:
What about splitting this class up into 2 classes, and using this naming scheme:
- `mb-md-40`: `margin-bottom` of 40px at `md` widths and above
- `mb-28`: `margin-bottom: of 28px

and then this way, these classes can be re-used if there are any other places that need either of these declarations.

angela-tran:
> What about splitting this class up into 2 classes, and using this naming scheme:
> 
>     * `mb-md-40`: `margin-bottom` of 40px at `md` widths and above
> 
>     * `mb-28`: `margin-bottom` of 28px
> 
> 
> and then this way, these classes can be re-used if there are any other places that need either of these declarations. And since the class names correspond mostly with Bootstrap's spacing schemes, it's easier to read for others.

I'm ok with this approach. The reusability is nice

thekaveman:
Oof, this seems like one of those times where there shouldn't have been a merge conflict... annoying

angela-tran:
Rebased to resolve conflicts. This is ready for re-review

thekaveman (approved):
Nice! I like the generalization of the helper classes, LGTM!

PR #197, opened 2023-06-14
Fix: Make About link jump to About section

### LINKED ISSUE #180: Make "About the project" navbar link jump to homepage project description
~MobiMart does not have a separate homepage link, it only uses the logo in the nav.~

See updated description in comment below: https://github.com/cal-itp/calitp.org/issues/180#issuecomment-1592022357

thekaveman:
Closes #180 

OK this definitely feels a little hacky so I am wide open to feedback if there is a better way.

Simply linking to the `#about` section causes the navbar to slightly obscure the header:

image

Based on a couple of SO posts:

* [Boostrap using fixed navbar and anchor tags to jump to sections](https://stackoverflow.com/a/17181529/453168) which talks about using padding and a corresponding negative margin to avoid the navbar overlap
* [How can I get a negative value of CSS variables](https://stackoverflow.com/a/48639938/453168) which talks about flipping a variable from positive to negative

I added a bit of CSS for this `section#about`, which (I think) gives the intended result:

**Desktop**

image

**Mobile**

image

### DIFF
--- a/src/_includes/header.html
+++ b/src/_includes/header.html
@@ -21,7 +21,7 @@
       <div class="navbar-collapse collapse justify-content-end" id="mainNav">
         <ul class="navbar-nav mb-0">
           <li class="nav-item">
-            <a class="h4 mb-0 nav-link d-block" href="/">About the project</a>
+            <a class="h4 mb-0 nav-link d-block" href="/#about">About the project</a>
           </li>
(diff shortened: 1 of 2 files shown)


esquared415:
@thekaveman Thank you! Looks good to me, hacks and all!

PR #199, opened 2023-06-15
Fix: Footer issues

machikoyasuda:
An assorted mix of footer-related fixes:

1. On your laptop, go to https://staging.calitp.org/press and click **GTFS** or **Benefits** - the footer comes off from the bottom of the page on desktop windows of a certain size. This is because there aren't enough Press release or Articles linked. In the meantime, this PR sets the `main.container` to be at least `calc(100vh - footer height)` so that the footer stays at the bottom.
2. The footer hover color wasn't working b/c of a CSS typo.
3. The spacing of the footer on mobile was a little off.

### DIFF
--- a/src/_includes/footer.html
+++ b/src/_includes/footer.html
@@ -1,28 +1,28 @@
 <section class="bg-dark navbar">
   <footer class="container">
     <nav class="col-12">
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Nice catch! LGTM

**Staging site**

image

**This PR**

image

PR #200, opened 2023-06-15
Fix: Home - Mobile: Initiatives

machikoyasuda:
Fixes another bug I noticed: 

- On mobile on the home page, the nav pills get stretched into 2 lines. Use `flex-shrink-0` to prevent the flex stretch.
- On mobile, the Initiatives div padding x should be 32px, not 48px.
- On mobile, the Initiatives's image has too much padding above, and not enough below. Balance it out.

## Screenshots

image

Prevent this situation:
image

Fixes:
image

### DIFF
--- a/src/_includes/initiatives.html
+++ b/src/_includes/initiatives.html
@@ -38,9 +38,9 @@
       tabindex="0">
       <div
         id="{{ initiative.id }}"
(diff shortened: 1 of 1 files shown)


angela-tran (approved):
This looks great @machikoyasuda !

PR #215, opened 2023-07-31
Fix: Initiatives section

### LINKED ISSUE #208: Update content within Initiatives section to better frame outbound link to MobiMart
Feedback from Marty at CalSTA was to better frame MobiMart within the context of our initiatives

Requirements: 
- Add line above button to contextualize what MobiMart is and how it connects to the initiative: "For more information about the products available to support this initiative, please visit the California Mobility Marketplace."
- Hide MobiMart button for Benefits and GTFS since we don't have products on MobiMart for Benefits or GTFS (We will add the button back for GTFS-RT when it is available towards the Fall)

machikoyasuda:
fix #208 

- hide Mobi Mart button for GTFS, Benefits
- add pre-button text for Contactless Payments

image
image
image

### DIFF
--- a/src/_data/initiatives.yml
+++ b/src/_data/initiatives.yml
@@ -16,6 +16,7 @@
       target='_blank'>Monterey-Salinas Transit</a>, Cal-ITP and partners like Visa
       demonstrated how a transit provider that has traditionally used cash and agency-specific
       fare cards can accept contactless bank card payments like any other merchant. "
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Tested this locally on desktop and mobile views 👍 

I like this approach, keeps the data structure more standard and the template logic simpler -- knowing that the button will likely come back at some point.

Issue #80, opened 2023-04-13
Implement new calitp.org redesign


mrose914:
High level requirements: 

- Shortens context about the program 
- Adds scrollable initiatives view 
- Adds Resources page 
- Adds Press page 
- Improves navigation to relevant content (for bitly site replacement)

Designs in progress here: https://www.figma.com/file/TfWk1iDHdlt7bSA3DjQuQv/Mobility-Marketplace?node-id=5299-23559&t=0sznFMu9eKaRxejn-0

thekaveman:
Remove assignment while blocked.

machikoyasuda:
Content approved by @esquared415 on April 30, 2023

Here: https://docs.google.com/document/d/1uLjCTC0LEjaDtOivWQp0OFM0ZVIjULnr_cZ6-QgAi0k/edit#heading=h.6xss7cceyx3k

segacy1:
Hi! I also wanted to flag matching the styles here to what's in mobimart (which is the current design). @mrose914 @angela-tran not sure if that's something that requires its own ticket or should be included in the description here!

thekaveman:
Reopening so we can close with the final PR from `staging` --> `main`

PR #223, opened 2023-07-31
Fix: broken GitHub Action

thekaveman:
We'll have to update again once https://github.com/cal-itp/calitp.org/issues/80 is closed and the redesign is launched

This is why [this action is failing](https://github.com/cal-itp/calitp.org/actions/workflows/dependabot-differ.yaml) on all the recent PRs


machikoyasuda (approved):
sounds good to me!

angela-tran (approved):

PR #226, opened 2023-08-03
Add Data Plans tag to Resources and Press

### LINKED ISSUE #225: Add 'data plans' as tag to both /resources and /press pages
Now that we have Data Plans available on MobiMart, we can and should add the option to tag content on calitp.org by this product/initiative.

machikoyasuda:
closes #225 

image
image

### DIFF
--- a/src/_data/press_tags.yml
+++ b/src/_data/press_tags.yml
@@ -4,3 +4,5 @@
   id: "benefits"
 - name: "GTFS"
   id: "gtfs"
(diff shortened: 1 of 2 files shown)


machikoyasuda:
@thekaveman @angela-tran this is ready for review.

in the next content update from Erika (which is not yet complete), there will be resources and press with the tag `Data Plans`. so this PR is in preparation for that.

PR #243, opened 2023-08-18
fix(resources): update resource pdf

machikoyasuda:
- new PDF
- new PDF name

Note: You have to convert DOCX'd Google Drive files into true Google Drive files, or else links will show up with the full DOCX file extension like this on LinkedIn and look sloppy:

image

This PDF was manually converted to a Google Drive file, banishing the DOCX file type, which changes the metadata title for the page to that of the file name, which shows up in two places: On Google Chrome like this - and also on LinkedIn.
image


machikoyasuda:
Confirmed non-sloppy file name:
image

thekaveman (approved):

PR #336, opened 2024-05-02
Feat: web-review team reviews content changes

thekaveman:
Adding @cal-itp/web-review (currently just @mrose914) to `CODEOWNERS` for specific files / directories where the content lives.

This will make it possible to automatically assign review for changes to these files. And means we won't have to wait for another engineer to review simple content updates. Checking with you @mrose914 is that is OK? We typically request your review anyway :sweat_smile:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,7 @@
 # default to web-admin team
 * @cal-itp/web-admin
+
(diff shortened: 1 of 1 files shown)


thekaveman:
Approved by @mrose914 in Slack: https://cal-itp.slack.com/archives/C018XMWT2FN/p1715887139066379

PR #363, opened 2024-07-15
Feat: Initiatives by permalink

### LINKED ISSUE #353: Home page: Permalinks for homepage Initiatives section
## Changes requested
- Make the Initiatives section linkable

machikoyasuda:
closes #353 

## How to test
- Go to http://localhost:4000/#pills-benefits or http://localhost:4000/#pills-benefits - make sure the page loads to the right tab and page height
image
- Click around other tabs to make sure the other tabs still also open (but the anchor tag in the URL will not change).

### DIFF
--- a/src/stylesheets/main.css
+++ b/src/stylesheets/main.css
@@ -305,6 +305,10 @@ footer a:hover {
   border-style: solid;
 }
 
+.scroll-link {
+  scroll-margin-top: 240px;
+}
+
 .black-on-white .nav-pills {
   --bs-nav-pills-link-active-color: var(--bs-white);
   --bs-nav-pills-link-active-bg: var(--bs-body-color);
--- a/src/_includes/initiatives.html
+++ b/src/_includes/initiatives.html
@@ -35,7 +35,7 @@
 <div class="tab-content" id="pills-tabContent">
   {% for initiative in site.data.initiatives %}
   <div
-    class="tab-pane fade {{ initiative.active }}"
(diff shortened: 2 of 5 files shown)


src/stylesheets/main.css
```diff
@@ -305,6 +305,10 @@ footer a:hover {
   border-style: solid;
 }
 
+.scroll-link {
```
machikoyasuda, line 308: This allows the anchor link to scroll an extra 240px (roughly double the height of the top nav), so that the anchor scrolls to here
image

instead of here:
image

In this scenario, the `Explore the intiatives` is hidden by the sticky nav.

thekaveman (changes requested):
This is really cool! Works great and the smooth scroll is nice.

One note on testing: this doesn't really work if you have already loaded the page, and then add the `#pills-whatever` to the URL and hit enter, or change to `#pills-whatever2`; I don't think this is a real use-case we need to be concerned about. It worked just fine if I entered the URL directly in the browser bar in a blank tab and hit enter (like what would happen if someone clicked a link).

**One small request:** can we swap the word `pills-` for something like `initiatives-` -- if these are going to be URLs that are shared, we want them to sort of make sense and not look weird :sweat_smile:

thekaveman (approved):
Very cool :sunglasses:

thekaveman:
cc @indexing we now have direct links to the Initiatives sections on `calitp.org`

PR #364, opened 2024-07-16
Partners

machikoyasuda:
closes #351 

## What this PR does
- [x] Home page: Add image logos
- [x] Partners page: Contact template
- [x] Partners page: Add photos, text, links
- [x] CSS: Add pulled quote text style
- [x] CSS: Add statistics text style
- [x] CSS: Add small attribution text style
- [x] CSS: All the spacings - Desktop
- [x] CSS: All the spacings - Mobile
- [x] Design: Add statistics box rhombus style

## Notes
- This PR changed the names for _all_ the CSS classes/SVG paths with rhombuses - that includes the homepage, Press, Resources and Partners page. So please review all the pages for any regressions (there shouldn't be though!).
- I'm starting this new thing of having a basic naming convention for images: `(name of page)-(type of image)-(slug)`, like `home-logo-calsta` or `partners-background-cdt`. Mobile only images have `--sm` appended at the end:
image

PR #322, opened 2024-03-31
Convert Minimum Accounts Memo to Markdown

allejo:
This PR converts the PDF to markdown but also touches the CSS to add support for footnotes and set a precedent on how to handle images with captions.

Closes #311

PR #380, opened 2024-07-25
Customer stories: Stories page and homepage update

### LINKED ISSUE #354: Customer stories page and home page stats
## Changes requested
- New page, customer stories
- Add this new page to the nav
- Add new stats section to the homepage, which links to the customer stories page

## New styles needed
- New yellow color
- Border radius'ed image class
- Stat font style
- Pulled quote style
- New grey skinny link style
- Photo caption style
- Bullet points
- Linked superscript
- Source

## Assets
- California map
- 4 photos

## New component
- Side by side stat component

## Preview deploy
Make a branch to deploy so this can be approved separately from the Partners page

machikoyasuda:
closes #354 

- [x] Nav: Add customer stories to nav
- [x] Stories page: Add all content and static TOC
- [x] Stories page: Impact section - Mobile
- [x] Stories page: Impact section - Large/Medium: Outer alignment
- [x] Stories page: Large/Medium: Fix the footer
- [x] Stories page: Impact section - Large/Medium: Inner alignment
- [x] Stories page: Figure out the fade in / out effect
- [x] Stories page: Add footnote links
- [x] Stories page: Anchor links
- [x] Home page: Add stats
- [x] Stories/Partners page: Footnote code

## How to test this PR
- Enter the page through an anchor tag: https://deploy-preview-380--cal-itp-website.netlify.app/customer-stories#stories-GTFS-VCTC https://deploy-preview-380--cal-itp-website.netlify.app/customer-stories#stories-benefits-MST
- Enter without anchor tag https://deploy-preview-380--cal-itp-website.netlify.app/customer-stories
- Click on the sidebar links - active area should be in blue
- After clicking on a sidebar link, click Back on the browser. The URL should change and the active title should match the browser URL.
- Check the stats on the homepage: https://deploy-preview-380--cal-itp-website.netlify.app
- Nav should have `Customer stories` highlighted

## Sidebar
- [x] Desktop: The sidebar stays sticky on scroll, but also goes behind the Impact section (and has the fade background)
- [x] Clicking / tabbing and entering on a link takes you to the section
- [x] Clicking / tabbing on a single link will make _both_ pairs of links colored in styling
- [x] If you enter the page with an anchor tag, like `#stories-GTFS-VCTD`, then the link pairs will be set to active and styled appropriately


machikoyasuda:
Hey @ohmegasquared, 

This page is not 100% done yet, but it's close. I have all the sections for both desktop and mobile aligned with the latest content and for the most part, have the font sizes, colors and margins/padding. There's some refinement with the inside of the `Our impact` section (font size, mobile inner stat alignment) that's not done yet, and the table of contents link don't work yet at all. But other than that, the major "bones" of the page are complete.

Here's how I approached coding the tricky `Our impact` (image and stats side-by-side/on-top design) section:
1. Get the mobile version working first. If all else fails, the mobile version will work.
2. Get the desktop widths of the image and the stat section working.
3. Get the desktop vertical alignment of the image and stat section working.
4. Get the way the table of contents goes underneath the stat section working. Get the table of contents stickyness working.
5. The last part I tried was the fade in/out linear gradient of the stat section. It's a little tricky because there's already a lot of code magic happening to make the rhombus-with-rounded-corners working. 

This is the way the stat section on desktop is constructed:

image

image

There is a background applied to the div highlighted in purple, that has the linear gradient: `linear-gradient(180deg, rgba(255, 255, 255, 0.25) 0.16%, #fff 30.54%)`. I got that code from Figma. It's basically saying, "make a vertical linear gradient that grades between #ffffff and 25% transparency". But, as you can tell from the screenshot, this means that the gradient also shows up here -- the 190px of overlap between the two sections:

image

I have hard-coded this 190px overlap by using a negative margin.

But what I think we need here is a gradient that's more complex than a linear gradient: a gradient that has the linear gradient on the whole rectangle, except, the left-hand half is entirely transparent. Trying to draw out what I mean in Google Spreadsheets:

image

CSS gradients can be made quite complex, but I don't think it can make a gradient of what I need. Instead of a CSS-coded gradient, can use an image of a gradient as a background instead of linear gradient code. 

Do you think you could make an image like this in Photoshop or Figma? 

There are some other approaches I can try after, but they're slightly hackier, so I want to try this gradient image approach first.

ohmegasquared:
Yes, here is an image gradient! Thanks for tackling this tricky problem! 

Let me know if you need a different size/change for the image

stories-bg-gradient

machikoyasuda:
@allejo Hey Vlad, I was wondering if you could give this a code review on two specific elements. Note that the copy in this PR is not 100% complete, so that's why, for example, the first footnote's source is not a link and it doesn't go anywhere:

1. Footnotes: This page has 2 footnotes. I copied the way you did it, code-wise - both CSS classes (`.footnotes`) and HTML elements (`<ol>`, `<li>`, `<sup>`), on #322.

image

I had some questions about how to approach the design of footnote sections overall:
- Does the `Footnotes` title have to be there? The design doesn't call for it. It's there now b/c I'm using the `footnotes` class you used.
- Is it an anti-pattern to have a link that is just titled, "Source"? Would it be better for the source to have a title and the link to use the title instead? Should this title use `<cite>`? 


2. Sidebar nav: This page has an interesting sidebar nav. When you click on a link, like `Benefits` - both `Benefits` and the `Monterey-Salinas Transit` links turn to active. There are 3 sections of text, and the sidebar is divided up into Topics and Agencies - each with 3 items. There is only one corresponding topic to one corresponding agency: GTFS-VCTC, Contactless-Glenn, Benefits-MST. Clicking on 1 section's title activates the link of the corresponding topic/agency pair. Do you see anything off with this, accessibility-wise or otherwise?

Thank you

src/_includes/footer.html
```diff
@@ -1,33 +1,32 @@
-<section class="bg-dark navbar d-print-none">
+<section class="{{ page.footer-class }} bg-dark navbar d-print-none">
```
machikoyasuda, line 1: I had to add this to get the footer to stay at the bottom on the Customer stories page. The way the `Our impact` section is absolutely positioned breaks everything below it.

src/stylesheets/main.css
```diff
@@ -433,17 +467,73 @@ footer a:hover {
...
+    z-index: 1025;
+    position: absolute;
+    left: 0;
+  }
+
+  .monterey-and-contact,
+  .absolute-footer {
+    position: relative;
+    top: 620px; /* Height of the absolutely-positioned Impact section */
+  }
```
machikoyasuda, line 518: This 620px is the height of the `Our impact` section. Since the `Our impact` section is absolutely positioned and placed vertically between other divs - and therefore taken out of the normal flow, the Monterey-Salinas Transit/Contact section and the footer will be 620 pixels up higher than it should be. This code will add 620px above each section so that the sections will be vertically positioned correctly.

machikoyasuda:
This PR is now ready for code and design review. Please review the code/design, and mark this PR as approved (or not) wih the GitHub button. Once this PR is approved, it will be merged into #364. After that, all the copy for both Partners and Customer Stories will be reviewed together.

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+  </div>
+</div>
+
+<section class="row">
+  <div class="offset-lg-1 col-lg-10">
+    <div class="row flex-lg-row justify-content-center justify-content-lg-start mt-lg-5">
+      <div class="col-11 order-lg-2 offset-lg-1 col-lg-3">
+        <div class="sticky-top stories-toc" tabindex="1">
+          <div class="d-none d-lg-flex flex-column gap-3 mb-md-4 mb-3">
+            <span class="small-caps border-0 bg-transparent fw-bold p-0">Topics</span>
```
thekaveman, line 65: I'm not sure what the best practice is for headings in a sidenav. I guess I feel like these could be `h2` but IDK.

* `Topics`
* `On this page`

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+          </figure>
+          <figure class="quote border-yellow-4 mt-3">
+            <blockquote>
+              “As a small agency who doesn’t have the resources to be specialists in these areas, we’re able to leverage Cal-ITP’s
+              support to get complicated tech projects done more quickly…or at all.”
+            </blockquote>
+            <figcaption class="ps-3">— Public Transit Director in Ventura County</figcaption>
+          </figure>
+          <h4 class="small-caps border-0 bg-transparent fw-bold text-primary p-0 mb-2" id="stories-GTFS-VCTC">GTFS</h4>
+          <h3>Ventura County Transportation Commission</h3>
```
thekaveman, line 115: I think semantically, these two headings need to be flipped, e.g. 

* `GTFS` / topic --> `h3`
* `VCTC` / agency --> `h4`

E.g. because VCTC is just one example of GTFS...

OR

Make this `GTFS` heading a `<span>` like the other topic headings. Either way, we should apply the same logic to all 3 sections.

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+            processes.
+          </p>
+        </div>
+
+        <div class="glenn pb-3">
+          <figure class="image pt-3 mt-3 pb-3">
+            <img alt="TBD" class="rounded-2 w-100" src="/images/stories-photo-glenn.jpg" />
+            <figcaption class="small text-end text-secondary pt-2">Courtesy of TBD</figcaption>
+          </figure>
+          <span class="small-caps border-0 bg-transparent fw-bold text-primary p-0 mb-2" id="stories-contactless-payments-GCTD"
```
thekaveman, line 146: Add `h3` and flip headings, or ignore.

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+                  </div>
+                  <figcaption class="d-block d-lg-none small text-secondary pt-2">Photo by Kyle Smith</figcaption>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+
+        <div class="monterey-and-contact">
+          <span class="small-caps border-0 bg-transparent fw-bold text-primary p-0 mb-2" id="stories-benefits-MST">Benefits</span>
```
thekaveman, line 239: Add `h3` and flip headings, or ignore.

thekaveman (changes requested):
* Small file rename suggestion
* Small code change request to the sidebar script
* Questions/comments on Stories page headings

src/stylesheets/main.css
machikoyasuda: It's just the height of the entire parent div (includes the image of the train and the rhombus plus the caption below image): 
image

So there isn't a great way to calculate it. I can turn it into a variable just to make it more readable, but it won't be calculated in anyway. This was the only way we can get an extra-wide div in the row _while_ preserving the sidebar table of contents position.

thekaveman: All good, thanks for the explanation!

ohmegasquared:
(missing '#stories' in homepage URL anchor links) 

Design approved on mobile and web 🚀

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+      <h1 class="mb-2">Our customers</h1>
+      <p class="mb-3 pb-3">
+        The California Integrated Travel Project (Cal-ITP) is working with transit agencies across the state (and country!) to
+        promote a modern and consistent transportation experience throughout California. Each and every transit agency in the
+        state can become a Cal-ITP customer. We provide free technical assistance and support to help agencies implement
+        Contactless Payments, Transit Benefits, GTFS and more.
+      </p>
+      <div class="row d-none d-lg-flex">
+        <div class="col-lg-8">
+          <span class="small-caps border-0 bg-transparent fw-bold p-0">Why it matters</span>
```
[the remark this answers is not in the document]
machikoyasuda: Changing these to H2.

src/customer-stories.html
machikoyasuda: So I tried changing the code to a structure where the `TOPIC` is an `h3` and the `Agency Name` is an `h4`. Then I ran the app in the `pa11y` https://pa11y.org/ test locally. And this actually returns an error, because then there's a Heading without any content under it. It's a Heading with another Heading under it. 

```
 • Error: Heading tag found with no content. Text that is not intended as a heading should not be marked up with heading tags.
   ├── WCAG2AA.Principle1.Guideline1_3.1_3_1.H42.2
   ├── #stories-contactless-payments-GCTD
   └── <h3 class="small-caps border-0 bg-transparent fw-bold text-primary p-0 mb-2" id="stories-contactless-payments-GCTD"></h3>
```

So I'm going to go with: `TOPIC` as the `h3` and `Agency Name` as a _span_. And each agency/topic section will be wrapped in its own `<section>`, and the whole thing will be wrapped in `<article>`.

src/customer-stories.html
machikoyasuda: Taking a cue from https://www.w3.org/WAI/tutorials/menus/ and going to mark this up with a `<nav>` and aria-labeled by, and also changing the `ul` to an `ol` - becuase the order matters.

machikoyasuda:
Posting the pa11y test results of current PR here:

```
pa11y http://localhost:4000/customer-stories

Welcome to Pa11y

 > Running Pa11y on URL http://localhost:4000/customer-stories

Results for URL: http://localhost:4000/customer-stories

 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.09:1. Recommendation:  change text colour to #73777d.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── html > body > main > section > div > div > article > section:nth-child(2) > figure:nth-child(1) > figcaption
   └── <figcaption class="small text-end text-secondary pt-2"> Courtesy of Vent...</figcaption>

 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.09:1. Recommendation:  change text colour to #73777d.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── html > body > main > section > div > div > article > section:nth-child(3) > figure:nth-child(1) > figcaption
   └── <figcaption class="small text-end text-secondary pt-2">Courtesy of TBD</figcaption>

 • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 3.09:1. Recommendation:  change text colour to #73777d.
   ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
   ├── html > body > main > section > div > div > article > section:nth-child(5) > figure:nth-child(8) > figcaption
   └── <figcaption class="small text-end text-secondary pt-2">Courtesy of Monterey-Salinas Tr...</figcaption>

3 Errors
```

machikoyasuda:
The Wednesday update:

@thekaveman This is ready for code re-review. 

I'd like to try to get this merged into #364 by the end of day today, if possible.

One reason is because I've now identified one accesssibility testing issue, that affects _both_ #364 and #354 -- the super small font style color is not dark enough. This color/size combo is used on photo captions on Customer Stories and on footnoes for the Partners page. I'd like to fix that in one combined PR, b/c the underlying code is the same. 

I have 2 small PRs that can go into Production, approved by @ohmegasquared, and ready for code review. These 2 PR's can then be rebased into the future combined PR.

As of right now, the copy is still not ready yet. So hopefully I can get the real copy in tomorrow when the copy is ready, in one PR.

machikoyasuda:
@allejo I ran https://deploy-preview-377--cal-itp-website.netlify.app/partners into Axe and got this: 
image

machikoyasuda:
Ugh - Not sure why but the Netlify deploy has not updated today. So the link is not trustworthy! Please test locally.

src/customer-stories.html
thekaveman: Wow, much trickier than I expected!

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+  </div>
+</div>
+
+<section class="row">
+  <div class="offset-lg-1 col-lg-10">
+    <div class="row flex-lg-row justify-content-center justify-content-lg-start mt-lg-5">
+      <div class="col-11 order-lg-2 offset-lg-1 col-lg-3">
+        <nav aria-labelledby="tocheader" class="sticky-top stories-toc" tabindex="1">
+          <div class="d-none d-lg-flex flex-column gap-3 mb-md-4 mb-3">
+            <h3 id="tocheader" class="small-caps border-0 bg-transparent fw-bold p-0">Topics</h3>
```
thekaveman, line 65: Sorry to be a pain, but I think these should be `<h2>` instead of `<h3>`, since they are the first headings on the page after the main `<h1>`.

I think leaving `<h2>Selected agency highlights</h2>` is fine.

src/scripts/sidebar.js
```diff
@@ -0,0 +1,25 @@
...
+
+  // find link pair and add active class to both links
+  function findPairAndAddActive(hash) {
+    linkPair = document.querySelectorAll("a[href='" + hash + "']");
+    linkPair.forEach((link) => link.classList.add("active"));
+  }
+
+  // call the function once on page load to check for the initial hash
+  checkStoriesHash();
+});
```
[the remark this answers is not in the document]
thekaveman: @machikoyasuda I didn't see this file get any updates, still seeing the behavior I described above.

src/customer-stories.html
```diff
@@ -0,0 +1,337 @@
...
+        <div class="col-lg-8">
+          <h2 class="small-caps border-0 bg-transparent fw-bold p-0">Why it matters</h2>
+        </div>
+        <div class="col-lg-4">
+          <h2 class="small-caps border-0 bg-transparent fw-bold p-0">Network of agencies</h2>
+        </div>
+      </div>
+      <div class="row">
+        <div class="col-lg-4 d-flex justify-content-center flex-column">
+          <span class="small-caps border-0 bg-transparent fw-bold p-0 d-lg-none">Why it matters</span>
```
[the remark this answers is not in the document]
thekaveman: Just one missing in the top stats box

allejo:
> Does the Footnotes title have to be there? The design doesn't call for it. It's there now on the customer stories page b/c I'm using the footnotes class you used.

Nope, "Footnotes" doesn't have to be there. Kramdown already generates that section with the appropriate* ARIA roles, so that should suffice.

\* It looks like Kramdown generates `<li>`s with `doc-endnote` roles, which is now deprecated, as mentioned in the [WAI-ARIA specification](https://www.w3.org/TR/dpub-aria-1.1/#doc-endnotes). This also appears to be a known issue in Kramdown (https://github.com/gettalong/kramdown/issues/729).

Doing some more research, [Pandoc removed the `doc-endnote` from their generated code and left the non-deprecated `doc-endnotes`](https://github.com/jgm/pandoc/commit/f9b2ca7465e53e78fa70115a803d403b5c71bc82) (notice the plural).

I'm not *too* concerned about this causing problems from an accessibility perspective, considering the accepted and valid `doc-endnotes` is being generated where it should be. I would either submit a PR to Kramdown to correct this behavior or just wait patiently until someone else does it.

> Is it an anti-pattern to have a link that is just titled, "Source"? Would it be better for the source to have a title and the link to use the title instead? Should this title use <cite>?

It is 100% bad practice to have a link just called "Source" or "Read more." Screen readers can extract all links from a page; this means that if an Assistive Technology (AT) user uses this feature, they would receive a dozen URLs with no context of where they came from (i.e., what the surrounding sentences were), and the only information they'd have is the name of the link, which would be "Source."

A better practice is for links to have enough context in the name to be useful standalone, e.g., the name of the page the link is referencing.

> Sidebar nav: This page has an interesting sidebar nav. When you click on a link, like Benefits - both Benefits and the Monterey-Salinas Transit links turn to active. There are 3 sections of text, and the sidebar is divided up into Topics and Agencies - each with 3 items. There is only one corresponding topic to one corresponding agency: GTFS-VCTC, Contactless-Glenn, Benefits-MST. Clicking on 1 section's title activates the link of the corresponding topic/agency pair. Do you see anything off with this, accessibility-wise or otherwise?

Hmm... I can't see this causing problems for AT users since the only things affected are the link's classes. If you were misusing the [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) attribute on those sidebar links, then that could cause confusion. It doesn't make sense to use that here since they're all links within the same page, so it wouldn't convey any extra meaning.

allejo:
Oops, I missed one!

> I have the divs ordered so that the sidebar gets focused on first before the actual content div. Does that seem logical?

Yes.

src/scripts/sidebar.js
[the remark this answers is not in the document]
machikoyasuda: Sorry! Now this is all in: https://github.com/cal-itp/calitp.org/pull/380/commits/c75ab7c04d9fab5594f3fbd69726e8b8319b8540 - tested it going forwards (clicking a Topic or Agency) and backwards (clicking Back)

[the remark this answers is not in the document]
thekaveman: This is much better! There is still one small change I think we need:

When the hash is removed entirely from the URL (e.g. land on the page normally, click a menu item, then click back button) -- we should deactivate all menu items. Otherwise the page scrolls back to the top, hash is empty, but the most recently selected menu item is still active.

Also, this is "Not a Big deal" :wink: but I don't think we need the `click` handler anymore, because the native behavior of clicking a link is to update the hash.

thekaveman (changes requested):
One more small fix for sidenav logic

machikoyasuda:
@thekaveman Added a final `else if`, for the going Back on a non-hashed url instance: https://github.com/cal-itp/calitp.org/pull/380/commits/68487afe0772b5e505738e6a6457e436e4b4fc61

src/scripts/sidebar.js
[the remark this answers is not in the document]
thekaveman: Nice!!!!!! :tada:

thekaveman (approved):
This looks soooooo nice! All the interactions feel very smooth. Incredible work!

PR #440, opened 2025-02-24
Add Python script for generating resource entries

### LINKED ISSUE #435: Exploration: From issue template to branches
I've been thinking of a little experiment to build. I want to try improving the process of adding resources to this repository. Right now, we're using Issue Templates to get the appropriate information from users. The format for issues created from templates are pretty standard. So as long as the user doesn't modify the issue contents too much, we can parse the issue body with something like `GrantBirki/issue-template-parser`.

```yaml
steps:
  - name: Parse Issue
    id: parse-issue
    uses: GrantBirki/issue-template-parser@vX.X.X
    with:
      body: ${{ github.event.issue.body }}

  - name: Output Issue JSON
    id: output-issue
    run: echo ${{ steps.issue-parser.outputs.json }}
```

This method isn't 100% reliable as manual edits or weird syntax could break this parsing method. However, if everyone plays nice, then this should be able to get the job done. I think it's a risk I'm willing to take.

Moving on, once we've parsed the issue body, the next task is to create a new branch with a new file inside of `src/_resources`. Next up, using the JSON from the parsed issue contents, we can now write a file with the filled in information.

```bash
title="${{ fromJSON(steps.parse-issue.outputs.json).title }}"
file_name=$(echo $title | tr '\s' '-')

cat <<EOF > src/_resources/case-study-$file_name.md
---
date: ...
title: $title
---
EOF
```

Now, this code could be written in Bash and let it be embedded in the workflow. Or we could also write this as a Python script in the repo and simply execute it from the workflow. 

Lastly, you'd be able to use something like [`EndBug/add-and-commit`](https://github.com/EndBug/add-and-commit) to add and commit this file to the new branch.

---

What are your thoughts on investing time into building something like this?

allejo:
There are a few moving parts to this workflow, so let's break them down.

## The Python Script

To reduce the number of dependencies we had to worry about in this workflow, I ported the logic found in [GrantBirki/issue-template-parser](https://github.com/GrantBirki/issue-template-parser) into Python so that all the logic fits in this one script. This gives us the added benefit of easily running it locally without the need for emulating a GitHub Actions environment. The debugging functionality and output will behave appropriately when the script is run in a CI environment vs locally.

This script expects a JSON object with the key of `body` as an argument fed via STDIN or as the first argument when running the script. This object shape is used as it is the output of,

```bash
gh issue view <issue num> --json body
```

## The Workflow

Next up: the workflow. The Python script itself has different ways that it can fail. When run in a CI environment, it will set step outputs that can be used to determine why it failed.

Supported "failure" scenarios:

- The expected keys (aka headers) were not found in the issue body. This can mean one of two things:
  1. Another, or no issue template was used so that means we can skip running this process
  2. The issue template is outdated or was modified manually where not all keys are up to date
- If an unapproved resource request is made (i.e., the "Approved?" checkbox is marked as "no," it will leave a comment on the issue saying this process was skipped.

## Example Runs

Here are two example runs of this workflow:

- Triggered by `workflow_dispatch`: https://github.com/allejo/calitp.org/pull/9
- Triggered by an issue being created: https://github.com/allejo/calitp.org/pull/10

---

<details>
<summary>Outdated as of 2025-03-02</summary>

Since I've only committed the Python script portion of this PR, I'll mark this as a draft for now. As I was planning this out and writing it, I ran into some things I wanted to get more opinions on.

## Should we download assets?

Downloading assets will fail if said files are not public. Some ideas on how to handle this situation:

1. Auto-respond to the issue and request that the file be made public
4. Use a service account to download files

## Security Concerns

Automatically downloading files from the internet is a very secure thing to do. How should we handle the situation?

1. Only download assets from a list of trusted GitHub usernames?
2. Don't make this into an action and leave it as a manual script to run
</details>

---

/cc @machikoyasuda @thekaveman

Closes #435

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Site analytics is tracked by Google Analytics, version 4. Ask an administrator t
 
 ## Resource Requests
 
-We have a semi-automated way of handling resource requests via a Python script combind with the [`gh`](http://cli.github.com/) command.
+We have a semi-automated way of handling resource requests via a Python script combined with the [`gh`](http://cli.github.com/) command.
 
 ```bash
 gh issue view <issue num> --json body | python .github/scripts/resource-template.py
--- a/README.md
+++ b/README.md
@@ -26,6 +26,14 @@ and will watch the site files for changes.
 
 Site analytics is tracked by Google Analytics, version 4. Ask an administrator to grant you access.
 
+## Resource Requests
+
+We have a semi-automated way of handling resource requests via a Python script combind with the [`gh`](http://cli.github.com/) command.
+
+```bash
+gh issue view <issue num> --json body | python .github/scripts/resource-template.py
+```
+
 ## License
 
 Content (including graphics, images, video, documents, and text) in this repository is licensed under [CC-BY 4.0][content-license].
--- a/.github/scripts/resource-template.py
+++ b/.github/scripts/resource-template.py
@@ -66,6 +66,10 @@ def parse_issue_body(body):
 
   return result
 
+#
+# Utility functions for CI/CD operations
+#
+
 def is_ci():
   return os.getenv('CI', 'false') == 'true'
 
@@ -80,6 +84,12 @@ def set_gha_output(name, value):
     print(f'{name}<<{delimiter}', file=fh)
     print(value, file=fh)
     print(delimiter, file=fh)
+
+def write_debug(message):
+  if os.getenv('RUNNER_DEBUG', '0') == '1':
+    print(f'::debug::{message}')
+  elif os.getenv('DEBUG', 'false') == 'true':
+    print(f'[DEBUG] {message}')
 
 #
 # Utility functions for I/O operations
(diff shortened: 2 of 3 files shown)


thekaveman:
@allejo hmm good point, we probably don't want to be doing that.

I don't think this is meant to fully replace the current workflow, just shortcut it a bunch and provide some early feedback for requestors (e.g. the preview link). 

Given that, maybe we just use the asset link as provided in the issue, as a first pass. An engineer still needs to review (and possibly update) the PR anyway. We can try this out a few times and see how we like it or what changes to make. Can always come back to update this process further.

README.md
```diff
@@ -26,6 +26,14 @@ and will watch the site files for changes.
 
 Site analytics is tracked by Google Analytics, version 4. Ask an administrator to grant you access.
 
+## Resource Requests
+
+We have a semi-automated way of handling resource requests via a Python script combind with the [`gh`](http://cli.github.com/) command.
```
machikoyasuda, line 31:
```suggestion
We have a semi-automated way of handling resource requests via a Python script combined with the [`gh`](http://cli.github.com/) command.
```

machikoyasuda:
Just a small typo review @allejo - Otherwise, this LGTM and am excited to test it out for new requests coming in.

angela-tran (approved):

PR #486, opened 2025-08-19
Some Slack notification changes

### LINKED ISSUE #474: Create alert for `deploy preview` and `build` command failure in #notify-websites channel on Cal-ITP Slack
As a product manager for Cal-ITP websites, I want to be alerted to any preview and deployment failures for content published on either Calitp.org or Mobimart as quickly as possible, in a public channel where others can be tagged in to assist. 

To ensure the dev and product team are properly alerted to any `deploy preview` and `build` command failures, set up an alert to go to the #notify-websites channel when the command fails to deploy in production.

allejo:
## Dedupe Press/Resource Slack Notifications

Reported by @thekaveman, on Slack, the "New Press/Resource Request" Slack notification was reporting the issue twice.

SCR-20250818-tnlo

This was fixed in 8e48028be38d1effc9dd7370420cc7370e249148.

> One workflow gets triggered per label that was added. For example, in
"content" and "resources". Our if statement checked with `contains`
against the `issue.labels.*.name` array, which included both labels
already since these labels were created alongside the issue; opposed to
being added individually.
>
> Instead, check that the label triggered this workflow is what we're
looking for.

## Report failed Netlify deployment previews to Slack

The Netlify App uses GitHub's pull request status API to communicate when Netlify's website checks fail, succeed, get canceled, etc. We can listen to these status updates and trigger a workflow when they happen.

```yaml
if: (github.event.state == 'failure' || github.event.state == 'error') && endsWith(github.event.context, '/deploy-preview')
```

The context of the status we care for is `netlify/cal-itp-website/deploy-preview`, but why make things so strict? I felt like checking the whole string could inadvertently cause problems if we rename the project on Netlify at some point.

closes #474

## Reporting to Slack on build failures is [a paid extra](https://docs.netlify.com/extend/install-and-use/setup-guides/netlify-app-for-slack/).

image

We can circumvent this by either hosting or using someone's service to accept POST APIs and translate that into a Slack message.

### DIFF
--- a/.github/workflows/notify-failed-previews.yml
+++ b/.github/workflows/notify-failed-previews.yml
@@ -1,0 +1,39 @@
+name: On Netlify Check Failure (PR only)
+
+on:
+  status:
+
+jobs:
+  notify-slack-on-failure:
+    if: (github.event.state == 'failure' || github.event.state == 'error') && endsWith(github.event.context, '/deploy-preview')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install GitHub CLI
+        uses: cli/cli@v2
+
+      - name: Check if commit is part of a PR
+        id: prcheck
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          prs=$(gh api \
+            -H "Accept: application/vnd.github+json" \
+            repos/${{ github.repository }}/commits/${{ github.event.sha }}/pulls)
+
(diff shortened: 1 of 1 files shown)


thekaveman:
I'm going to rebase this PR to just include the fix for duplicate messages. We're going to drop the idea of failed build notifications as we don't have budget for extra paid features right now.

Issue #498, opened 2025-10-08
There is an empty page that should either not exist or redirect to proper resources


vevetron:
https://www.calitp.org/resources/case-study-cal-itp-statewide-fare-guidelines-toolkit ->

https://resources.calitp.org/calitp/Cal-ITP.Statewide.Fare.Guidelines.Toolkit.pdf

PR #508, opened 2025-11-03
chore: Standardize behavior for Resources and Press that don't render an internal page

Scotchester:
Follow-up to #504; part of finishing #498.

I had said,

> Will also need to ask Google to re-crawl these pages after this change is deployed in order to get them removed from the index. See: https://developers.google.com/search/docs/crawling-indexing/block-indexing#debugging-noindex-issues

but digging a little deeper now, the recommended way to do that in bulk is to update the `<lastmod>` tag on the sitemap entries for these pages, then ask Google to proactively check out the sitemap again.

This PR manually sets the last modified date on the placeholder resource pages to this morning.

---

**Update:**

This PR also now standardizes the behavior for Resources and Press items that don't render an internal page.
- The placeholder page for Resources with an `asset` property will now redirect to that asset value, as was already the case for Press items with an `external` property.
- External press item ghost pages also now have the `noindex` directive in their `<head>`.

### DIFF
--- a/src/_includes/meta.html
+++ b/src/_includes/meta.html
@@ -41,6 +41,6 @@
 <meta name="twitter:domain" content="{{ site.domain }}" />
 
 <link rel="canonical" href="{{ page.url | absolute_url }}" />
-{% if page.asset -%}
+{% if page.asset or page.external -%}
   <meta name="robots" content="noindex">
 {%- endif -%}
--- a/src/_layouts/resource.html
+++ b/src/_layouts/resource.html
@@ -2,6 +2,14 @@
 layout: default
 ---
 
+{% if page.asset %}
+  {% assign url = page.asset %}
+  {% unless page.asset contains "https://" %}
+    {% assign url = "/assets/" | append: url %}
+  {% endunless %}
+  {% include script_redirect.html location=url %}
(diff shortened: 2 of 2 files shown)


jgravois (approved):
this seems reasonable enough to me. 

i verified that the urls of the assets themselves already have fresh `<lastmod/>` values [on production](https://www.calitp.org/sitemap.xml), but many of the pages touched here are quite stale.

```xml
<url>
  <loc>https://www.calitp.org/resources/fact-sheet-gtfs-feed</loc>
  <lastmod>2021-05-01T17:00:00-07:00</lastmod>
</url>
```

since the problem is specific to DuckDuckGo, they are the ones we'll need to ask to reindex the site, correct?

Scotchester:
> since the problem is specific to DuckDuckGo, they are the ones we'll need to ask to reindex the site, correct?

🤦 I somehow missed that it was DuckDuckGo and not Google. Will look into how to ask them to reindex.

Also moving back to draft following discussion with Kegan this morning where we agreed to make the Resources and Press sections behave the same way.

thekaveman (approved):
LGTM 👍

jgravois (approved):
🎸

PR #533, opened 2025-11-25
docs: enable mkdocs netlify build previews

### LINKED ISSUE #524: Enable docs previews for Pull Requests
When a PR comes in that updates the contents of the docs site, or docs configuration, we should build and deploy a preview of the docs site with those changes.

And then post a comment on the PR with a link to the preview site.

See the [workflow in Benefits](https://github.com/cal-itp/benefits/blob/main/.github/workflows/mkdocs.yml#L27) for an example of how to set this up.

jgravois:
resolves #524

view it live: https://calitp-website-533--cal-itp-previews.netlify.app/

~before we can verify this PR does what it intends to, we have to grant this repo access to the two organization-wide secrets below:~

* ~`NETLIFY_PREVIEW_APP_SITE_ID`~
* ~`NETLIFY_AUTH_TOKEN`~

while i was in there i also did a little cleanup to further simplify the docker setup and prepare to host the site on https://doc.calitp.org/calitp.org (as soon as it has content worth publishing).


.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,6 @@
   "dockerComposeFile": "compose.yml",
   "service": "site",
   "runServices": ["site", "docs"],
-  "otherPortsAttributes": { "onAutoForward": "ignore" },
```
jgravois: this one doesn't appear to be needed.

.github/workflows/mkdocs.yml
```diff
@@ -0,0 +1,68 @@
...
+          cache: pip
+          cache-dependency-path: "docs/requirements.txt"
+
+      - name: Build MkDocs website
+        run: |
+          pip install -r docs/requirements.txt
+          mkdocs build
+
+      - name: Install Netlify CLI
+        run: npm install --location=global netlify-cli@23
```
jgravois: this is 6 major versions ahead of what we're using in benefits, but the CLI appears to be unchanged for the flags we're referencing.

mkdocs.yml
```diff
@@ -26,7 +26,6 @@ plugins:
 
 extra_css:
   - styles/theme.css
-  - https://use.fontawesome.com/releases/v5.13.0/css/all.css
```
jgravois: we can add this back if/when we start actually leaning on FontAwesome

.github/workflows/mkdocs.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: steps.find-comment.outputs.comment-id == ''
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const hostnameSuffix = "cal-itp-previews.netlify.app"
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: `Preview url: https://${context.repo.repo}-${{ github.event.number }}--${hostnameSuffix}`,
```
jgravois: i expect it should be fine to host this site at https://docs.calitp.org/calitp.org when we're ready, but i expect it _would_ be problematic as a preview url...

https://calitp.org-PR--cal-itp-previews.netlify.app/

in 2bd3249 i'm proposing that we hardcode a safe alias in this workflow, but if y'all would rather rename the repo to 'calitp-dot-org' or 'calitp-website', i can't say i have a problem with that either.

.devcontainer/compose.yml
```diff
@@ -15,6 +15,6 @@ services:
     entrypoint: mkdocs
     command: serve --dev-addr "0.0.0.0:8001"
     ports:
-      - "8001:8001"
+      - 8001:8001
```
jgravois: this is just me being neurotic and preferring numbers since thats what they show in the docker documentation.

PR #536, opened 2025-12-02
refactor: Break CSS out into multiple files

### LINKED ISSUE #529: Refactor `main.css` into multiple files
`main.css` isn't _huge_ right now, but @jgravois and I think it could be nice to break it into a few chunks, if for no other reason than it'd help us lock in an solid understanding of how the styling is currently architected.

Scotchester:
ℹ️ **PR is to base branch `feature/dsdl`** ℹ️ 

---

Closes #529 

This might be overkill, and it's mildly annoying to test it comprehensively, so if we decide not to merge it, I'm totally fine with that. It was indeed a useful exercise in getting acquainted with what all we've got going on in this site, though.

This PR doesn't do anything directly relating to the DSDL implementation, but here's what it does do:

- Breaks out universal items into new files that get imported at the top of `main.css`:
  - variables (Bootstrap overrides and fully custom)
  - the component I'm calling "clipped containers" (anything involving `clipped-` classes)
  - footnotes
  - nav pills
  - the main navbar
  - print styles
- Breaks out page-specific styles into their own files, loaded via new `stylesheet` front matter that, if present, adds a new `<link>` to the `<head>`
- Core styles (heading styles, links, figures, etc.) and utilities remain in `main.css`

To test, the simplest thing is to load up the preview and compare it to the live site. You should not see any visible changes.


src/_includes/styles.html
```diff
@@ -10,3 +10,4 @@
       integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH"
       crossorigin="anonymous" />
 <link rel="stylesheet" href="/stylesheets/main.css" />
+{% if page.stylesheet %}<link rel="stylesheet" href="/stylesheets/{{ page.stylesheet }}" />{% endif %}
```
jgravois: in a future iteration of this we can make `main.css` itself optional and load `dsdl.css` selectively here too 😎

jgravois (approved):
i wouldn't have thought to approach the problem this way, but now that i've chewed on it i definitely think it makes sense to reorganize the closet before we start stuffing even more clothes in there.

nice work and i'm glad you got more acquainted with the overall existing architecture in the process.

PR #542, opened 2025-12-05
Update new_press.yml

leelas-pm:
Taking out reference to Erika in the approval field and replacing with language to match the Resource request form.


jgravois (approved):
🌈

PR #553, opened 2025-12-11
feat: Add spacing tokens and demo

### LINKED ISSUE #550: DSDL: Implement spacing utilities
Got about 5 minutes into #727 before realizing that we really need to get our new 8px-based spacing system implemented. I needed a 32px spacer right away, and your options in Bootstrap go from 24px (`pt-4`) straight to 32px (`pt-5`).

I'll bounce back to the DSDL code in calitp.org to add spacing utilities based on [the list provided by @cmajel in Figma](https://www.figma.com/design/IuVNsLccmvumsUzW7IC5BF/Digital-Services-Design-Language?node-id=1421-2&t=ACLDpjyS2bZ9a5pw-0)

Preview:

Diagram of spacing system

Scotchester:
Closes #550 

This PR adds spacing tokens to the DSDL CSS and adds a chart of them to the [demo page](https://deploy-preview-553--cal-itp-website.netlify.app/dsdl).

Review commit-by-commit to avoid seeing some minor cleanup until last.

### For discussion at some point

We currently use [Bootstrap's spacing utility classes](https://getbootstrap.com/docs/5.3/utilities/spacing/) extensively throughout our projects. This new 8px-based spacing system creates a tension with those in projects where we do not use Sass, because [there is not a CSS variable for overriding their base `$spacer`](https://getbootstrap.com/docs/5.3/customize/css-variables/) and [adding the additional increments we want would require overriding a Sass map](https://getbootstrap.com/docs/5.3/utilities/spacing/#sass-maps).

Generating our own custom spacing utilities would require a massive amount of work to cover all of the same functionality that Bootstrap's offer (margin + padding * six directional options * 14 increments).

The two options I see are:

1. Convert to Sass everywhere so we can override Bootstrap's Sass.
2. Lean away from using spacing utility classes and apply them via other CSS selectors.

(not necessarily in order of my preference – I'm not 100% sure yet which I think is best)

### DIFF
--- a/src/stylesheets/dsdl/tokens.css
+++ b/src/stylesheets/dsdl/tokens.css
@@ -60,8 +60,6 @@
 /* SPACING */
 
 :root {
-  --spacing-1px: 1px;  /* Non-relative unit for when we don't want a single pixel stroke to scale */
-
   --spacing-base: calc(8 / var(--text-base) * 1rem);  /* 8px → 0.5rem */
   --spacing-05:   calc(var(--spacing-base) * 0.5);  /* 0.25rem (4px)   */
   --spacing-1:    calc(var(--spacing-base) * 1  );  /* 0.5rem  (8px)   */
--- a/src/dsdl.html
+++ b/src/dsdl.html
@@ -337,16 +337,6 @@ stylesheet: dsdl-demo.css
     <tbody>
       <tr>
         <td>
-          <div class="spacing-sample" style="height: var(--spacing-1px);">
-            <span class="visually-hidden">A box illustrating the size of something using spacing 1px</span>
-          </div>
-        </td>
-        <td>--spacing-1px</td>
-        <td>1px</td>
-        <td>N/A</td>
-      </tr>
-      <tr>
-        <td>
           <div class="spacing-sample" style="height: var(--spacing-05);">
             <span class="visually-hidden">A box illustrating the size of something using spacing 05 (half)</span>
           </div>
@@ -488,5 +478,7 @@ stylesheet: dsdl-demo.css
     </tbody>
     <caption class="visually-hidden">Chart of our spacing variables with visual reference</caption>
   </table>
+
+  <p>Note: For lines we want to be exactly 1 logical pixel without scaling, we use a <code>1px</code> value directly.</p>
 
 </div>
--- a/src/stylesheets/dsdl-demo.css
+++ b/src/stylesheets/dsdl-demo.css
@@ -41,10 +41,13 @@
   text-align: center;
 }
 
+.dsdl-demo table {
+  margin: var(--spacing-2) 0;
+}
 .dsdl-demo table th,
 .dsdl-demo table td {
   padding: var(--spacing-1);
-  border: var(--spacing-1px) solid var(--dsdl-gray-20);
+  border: 1px solid var(--dsdl-gray-20);
 }
 .dsdl-demo table th {
   background: var(--calitp-brand-blue);


src/stylesheets/dsdl/tokens.css
```diff
@@ -49,7 +50,30 @@
...
+  --calitp-brand-blue:   var(--dsdl-blue-70);    /* #045b86 */
+  --calitp-brand-cyan:   var(--dsdl-cyan-40);    /* #38abd0 */
   --calitp-brand-yellow: var(--dsdl-yellow-30);  /* #febd28 */
 }
+
+
+/* SPACING */
+
+:root {
+  --spacing-1px: 1px;  /* Non-relative unit for when we don't want a single pixel stroke to scale */
```
jgravois: not sure i see the benefit of canonizing this as a CSS variable, but i'm okay with it.

src/stylesheets/dsdl/tokens.css
```diff
@@ -49,7 +50,30 @@
...
+
+  --spacing-base: 8;  /* Calculated pixel sizes based on this base →→→→→→→→→→→→→ ↓↓↓↓↓ */
+  --spacing-05:   calc(var(--spacing-base) * 0.5 / var(--text-base) * 1rem);  /* 4px   */
```
jgravois: i could be missing something but defining a `px` base and calculating final values in `rem` makes these conversions feel unnecessarily convoluted.

```css
// alternatives
:root {
  --spacing-base: 8px;
  --spacing-05: calc(var(--spacing-base) * 0.5); /* 4px */

  --spacing-base: 0.5rem;
  --spacing-05: calc(var(--spacing-base) * 0.5); /* 4px */
}
```

jgravois (commented):
this is already looking pretty promising.

> I'm not 100% sure yet which I think is best

me neither tbh. personally i don't see much/any downside to just using Sass to:
* customize bootstrap to give us all the 8px spacing helpers we need
* compile our own CSS back into a single file
* take advantage of its own nesting

that said, i can live without all that. if we're going to stick to pure CSS my instinct would be to start by defining only the missing spacing increments that we have immediate use for (24px, 80px).

we can try a couple things as we start to lay out mobimart pages and see what feels nicest.

src/stylesheets/dsdl/tokens.css
Scotchester: I did it to stay in sync with how Christine laid out the spec, but I think you're right that we should ditch it. I agree with the general principle (hence not adding variables for the different font weights before).

src/stylesheets/dsdl/tokens.css
Scotchester: I agree it's convoluted.

The first alternative doesn't work if we want the result units to be `rem`.

The second alternative I don't like exactly as written because when we base things off of pixel values (as in the text sizing above), I like seeing that number. It could be added via comment, though, OR we could do something like this:

```css
:root {
  --spacing-base: calc(8 / var(--text-base) * 1em; /* 8px → 0.5rem */
  --spacing-05: calc(var(--spacing-base) * 0.5); /* 0.25rem (4px) */
}
```

Scotchester:
> Review commit-by-commit to avoid seeing some minor cleanup until last.

I realize now that I didn't include the new `--text-base` var in the first commit, which is required for the spacing vars to work. Will fix that with a rebase before merging.

src/stylesheets/dsdl/tokens.css
Scotchester: Hrmmm. I wonder if that's just how Chrome displays it? I suppose the true test would be does it scale if you change the default text size in the browser prefs. I can double-check this.

jgravois: i also learned that github.com fails to honor non-default browser text sizes completely. i was surprised by this because they've made loads of accessibility improvements over the last few years.

src/stylesheets/dsdl/core.css
```diff
@@ -113,7 +113,7 @@ small,
   font-weight: 400;
 }
 .text-code {
-  font-family: var(--monospace-font-stack);
+  font-family: var(--dsdl-monospace-font-stack);
```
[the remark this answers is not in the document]
jgravois: since you're proposing this in #554, lets revert the change here.

Scotchester:
@jgravois Rebased and pushed an update addressing the above comments.

PR #554, opened 2025-12-11
feat(dsdl): Reintroduce separate heading and body font stacks

Scotchester:
ℹ️ **PR is to base branch `feature/dsdl`** ℹ️ 

---

Per feedback from @cmajel in #547

PR #555, opened 2025-12-16
chore: Cap the weight of Noto Sans at 600

Scotchester:
Follow up to comments in #554 

- Caps the weight of Noto Sans at 600 (https://github.com/cal-itp/calitp.org/pull/554#discussion_r2612164507)
- Removes superfluous heading styles (https://github.com/cal-itp/calitp.org/pull/554#discussion_r2612164689)

I did some testing to show that the weight will not be artificially made bolder if attempted:

Table of all weight/style combinations for Noto Sans

(tested in Firefox, Chrome, and Safari on macOS)

### DIFF
--- a/src/stylesheets/dsdl/core.css
+++ b/src/stylesheets/dsdl/core.css
@@ -10,15 +10,6 @@ body {
   line-height: var(--line-height-normal);
 }
 
-h1, .h1,
-h2, .h2,
-h3, .h3,
-h4, .h4,
-h5, .h5,
-h6, .h6 {
-  font-family: var(--dsdl-heading-font-stack);
-}
-
 .text-body-s {
   font-size: var(--text-xs);
   line-height: var(--line-height-loose);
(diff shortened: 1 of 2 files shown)


src/stylesheets/dsdl/core.css
```diff
@@ -10,15 +10,6 @@ body {
...
 
-h1, .h1,
-h2, .h2,
-h3, .h3,
-h4, .h4,
-h5, .h5,
-h6, .h6 {
-  font-family: var(--dsdl-heading-font-stack);
-}
-
```
lalver1: I've been learning a lot from following these PRs 😄 

Now this removal makes sense to me, since we have the same definitions below 👍 I was just thinking that maybe we can also move the `h5` block below so that it follows the order 6, 5, 4, etc. Not a big deal but just noticed that.

lalver1 (commented):
Just leaving a minor suggestion and a question below, thanks!

jgravois (approved):
as always, less is more!

Issue #516, opened 2025-11-10
Create DSDL documentation for calitp.org implementation


leelas-pm:
Depends on #522

leelas-pm:
> [@leelas-pm](https://github.com/leelas-pm) sounds good 👍 we can keep this one as a task/story, separate from setting up the docs site itself.

Ok, sounds good! I'll move this under the calitp.org DSDL epic then. Thank you!

thekaveman:
FYI @Scotchester @jgravois I added this to our Jan Sprint as Stretch, would be nice to get some more docs up since we're going to reference this docs site in the 012M wrap-up report.

PR #562, opened 2026-02-02
docs: Add DSDL explanation and reference pages

Scotchester:
Part of #516

---

This PR currently adds a DSDL page to the Explanation section of [the Cal-ITP Websites docs](https://docs.calitp.org/calitp.org/).

Following the [Diataxis](https://diataxis.fr/) framework that we're now using to structure our docs, my idea was to put the general philosophy and high-level design decisions into the Explanation section, put general instructions for using it into a Guides page, and then document the implementation details in the Reference section.

Given my experience using the existing DSDL code here (incorporating it into examples in MkDocs without trying to completely re-theme MkDocs – see `overrides/main.css` and some of the comments in `theme.css`), I think we'll want to do a little bit more work on how the DSDL is currently implemented before doing those pages.

We can maybe eliminate [the existing demo page](https://www.calitp.org/dsdl) once we finish building out all these parts of the docs. On the other hand, it may prove useful to continue to have a place to demo that is free of the baggage of MkDocs.

I think this could probably be merged as-is, but opening as a draft until we get @machikoyasuda reoriented and reintegrated.

### DIFF
--- a/docs/reference/dsdl/.pages
+++ b/docs/reference/dsdl/.pages
@@ -2,3 +2,4 @@ nav:
   - Colors: reference/dsdl/colors
   - Typography: reference/dsdl/typography
   - Spacing: reference/dsdl/spacing
+  - Utilities: reference/dsdl/utilities
--- a/docs/reference/dsdl/utilities.md
+++ b/docs/reference/dsdl/utilities.md
@@ -1,0 +1,53 @@
+# Digital Services Design Language utility class reference
+
+Like many design systems and frameworks, the DSDL offers some utility classes
+that can be applied when you want to ensure a specific value in a certain situation
+(enforced with the use of `!important`).
+
+
+## Font families
+
+```css
+.font-heading { font-family: var(--dsdl-heading-font-stack) !important; }
+.font-body { font-family: var(--dsdl-body-font-stack) !important; }
+.font-monospace { font-family: var(--dsdl-monospace-font-stack) !important; }
+```
+
+## Font sizes
+
+```css
+.font-size-xxs { font-size: var(--text-xxs) !important; }
+.font-size-xs { font-size: var(--text-xs) !important; }
+.font-size-s { font-size: var(--text-s) !important; }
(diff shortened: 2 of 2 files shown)


jgravois (commented):
i know this is still in draft, so if my question is premature, just ignore me.

docs/explanation/dsdl.md
```diff
@@ -0,0 +1,123 @@
...
+End users will benefit from a consistent feeling across each product, increasing usability through predictability and building trust in the Cal-ITP brand.
+
+For our own efforts in maintaining these products, using a unified structure of color, spacing, type, and components will reduce error and increase speed for both design and dev, while including flexibility to meet specific product needs.
+
+For example, there is a comprehensive color palette from which each product can draw on different parts of it and feel unified while maintaining their own individual identity. Similarly, we have base styles and sizes for typography, but can use different font stacks to adapt to each product.
+
+
+## Typography
+
+To maintain flexibility across products, we use a set scale with multiple typeface options for brand typefaces (larger, stylized typefaces used for emphasis and headlines, or "display" copy) and base typefaces (highly legible typefaces used for body copy, smaller captions and labels, etc.).
```
jgravois: maybe a _little_ clearer?

```suggestion
For example, a comprehensive color palette is included. Each product can draw on different parts of it and feel unified while maintaining its own individual identity.


## Typography

Similarly, we have base styles and sizes for typography, but can use different font stacks to adapt to each product.

To maintain flexibility, we use a set scale with multiple typeface options for brand typefaces (larger, stylized typefaces used for emphasis and headlines, or "display" copy) and base typefaces (highly legible typefaces used for body copy, smaller captions and labels, etc.).
```

jgravois (commented):
CIL for a couple questions. overall this looks rad.

Scotchester:
> Looks great. One open question that John had: https://github.com/cal-itp/calitp.org/pull/562/changes#r2756249071

Whoops, missed that one, thanks. File has been removed.

PR #477, opened 2025-07-16
Standardize Ruby version

allejo:
Reported in #469, Netlify inadvertently broke our builds because we never locked in a Ruby version. To address this, I temporarily set the `RUBY_VERSION` environment variable to `3.1` on Netlify, which resolved the issue for a week at most 🫠 Now, [Ruby 3.1 download URLs are 404'ing](https://github.com/cal-itp/mobility-marketplace/pull/656#issuecomment-3070111298), breaking the build pipeline again.

Round 2: Let's introduce a proper fix. This PR introduces a new `.ruby-version` file to the root of the repository, allowing Netlify to use it to install the correct version of Ruby during the build process. According to [Netlify's documentation regarding dependencies](https://docs.netlify.com/build/configure-builds/manage-dependencies/#ruby), the `.ruby-version` file must **NOT** have a trailing newline.

> The .ruby-version file must include the version number only: x.y.z, with no trailing newline.

However, in this PR's deploy previews, it appears that Netlify _can_ handle trailing newlines, so I'll include it since that's our de facto for the rest of our files.

Secondly, we want to ensure that our dev container also respects the `.ruby-version` file. I've modified the Dockerfile to install [rvm](https://rvm.io/), and then proceed to respect the version within this file.

Lastly, to fix the underlying problem going on here, I've added `bigdecimal` to the Gemfile as a dependency (see [my comment explaining why adding this dependency is necessary](https://github.com/cal-itp/calitp.org/issues/469#issuecomment-3018127982)).

closes #469

PR #589, opened 2026-03-31
chore: configure node (and 11ty) in devcontainer

jgravois:
closes #584

this PR goes a smidge further than https://github.com/cal-itp/mobility-marketplace/pull/749

* updates the python container image to use 3.12
* makes Node.js 24x available in the DevContainer (via mise)
* lays down 11ty

once you've rebuilt the DevContainer:

```bash
$ node -v
v24.14.1

$ npm list
calitp.org@2.0.0 /usr/src/calitp
├── @11ty/eleventy@3.1.5
└── @quasibit/eleventy-plugin-sitemap@2.2.0

$ npm start
Wrote 0 files in 0.12 seconds (v3.1.5)
[11ty] Eleventy Fatal Error (CLI):
```

i've configured https://11ty--cal-itp-website.netlify.app/ as an alternate production branch for when we get further along with the migration, but this piece is safe to land as-is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 .jekyll-*
 .bundle
+node_modules
 .netlify
 _site
 .DS_Store
 site
+# netlify uses mise to manage ruby, locally we use rvm
+mise.local.toml
--- a/.cspell/dev-dictionary.txt
+++ b/.cspell/dev-dictionary.txt
@@ -59,9 +59,11 @@ noteref
 Noto
 numList
 peter-evans
+pipefail
 prcheck
 previtem
 pymdownx
+quasibit
 recv
 redhat
 rgb
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,17 @@
-FROM python:3.10
+FROM python:3.12
 
 WORKDIR /usr/src/calitp
+
+# Install mise (to manage Node.js) https://mise.jdx.dev/mise-cookbook/docker.html#docker-image-with-mise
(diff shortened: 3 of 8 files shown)


angela-tran (approved):
I went back through older `mise`-related PRs (https://github.com/cal-itp/calitp.org/pull/477, #507) to better understand the current set-up of local -> `rvm` / Netlify -> `mise` and through the comments on [MobiMart's #749](https://github.com/cal-itp/mobility-marketplace/pull/749) (the counterpart to this PR) to better follow these changes. I'm all caught up here, and this looks good!

I was able to rebuild my devcontainer with this branch and run `npm`/`npx` too.

.gitignore
```diff
@@ -1,6 +1,9 @@
 .jekyll-*
 .bundle
+node_modules
 .netlify
 _site
 .DS_Store
 site
+# netlify uses mise to manage ruby, locally we use rvm
+mise.local.toml
```
angela-tran, line 9: _(non-blocking thought I had after submitting my review 😅)_

At first I was thinking it'd be helpful to document the explanation in https://github.com/cal-itp/mobility-marketplace/pull/749#discussion_r2636338143 about this `mise.local.toml` file in a place that's more permanent, but then I realized this is really only needed while we have both Ruby and Node in use. 

So seems like maybe we'd even remove this entry after cutting over to 11ty and removing Ruby/Jekyll completely?

jgravois: exactly. its just an (optional) stopgap measure to avoid downloading ruby twice until we say goodbye to jekyll.

PR #590, opened 2026-04-03
chore: convert static site builder from Jekyll to Eleventy

angela-tran:
closes #581 

## Testing steps

The easiest/best way to test is to open the two sites side by side and click back and forth.

https://www.calitp.org
https://11ty--cal-itp-website.netlify.app

in order to merge it we all need to be confident that it doesn't introduce any visual regressions, any adverse changes to SEO tags and that the sitemap.xml, robots.txt, redirects, google analytics etc. all look and function the way they should.

## Known differences
* `/assets/*.pdf` are no longer included in [/sitemap.xml](https://github.com/cal-itp/calitp.org/pull/614#issuecomment-4329182012)
* `datetime` now displays dates only (see https://github.com/cal-itp/calitp.org/pull/618#discussion_r3156721698 for more info)
* a thumbnail image is now displayed in shareable links
* `<meta property="og:description">` values now display frontmatter leads and intros (related: https://github.com/cal-itp/calitp.org/pull/618#discussion_r3157840138)

prior art: https://github.com/cal-itp/mobility-marketplace/pull/818


jgravois:
these are the regressions that bit us in MobiMart after our initial 11ty launch.

1. https://github.com/cal-itp/mobility-marketplace/issues/839
(we've definitely already got this one [patched up](https://github.com/cal-itp/calitp.org/blob/2bb6f89bbde96f2731231cf8f3017ffb1c10187d/src/404.html#L3))
1. https://github.com/cal-itp/mobility-marketplace/pull/835 
(we shouldn't have any trouble with missing files because we're already passing through anything/everything in [`/assets/`](https://github.com/cal-itp/calitp.org/blob/11ty/eleventy.config.js#L28). as for the malformed HTML, i'm 99% sure we don't have any raw HTML in front-matter anywhere in this site.)

cmajel:
It looks like this resources block on the homepage is missing from the 11ty version, for all the categories.

prod:
Screenshot 2026-05-04 at 1 30 36 PM
Screenshot 2026-05-04 at 1 35 20 PM


11ty:
Screenshot 2026-05-04 at 1 30 50 PM
Screenshot 2026-05-04 at 1 34 37 PM

src/index.liquid
```diff
@@ -1,160 +1,192 @@
...
+      </ol>
+    </div>
+  </div>
+</section>
+
+<picture class="d-none d-md-block">
+  <img
+    class="w-100 my-md-5 py-md-5"
+    src="images/home-divider-2.png"
+    alt="Decorative element with dots and dashes, meant to resemble a transit map"
```
cmajel, line 143: This can be separate from this PR (because it's also on prod), but we can remove alt text for these decorative images for a better screenreader experience.

cmajel:
When clicking through on press pages that go to external sites, I noticed the 11ty version would redirect slower, and and shows a flash of a "redirecting" page, but this does not happen on prod (from what I can tell). Is that intended? 

Screenshot 2026-05-04 at 1 52 23 PM

src/_includes/contact.html
```diff
@@ -44,18 +47,23 @@ <h3 class="text-white d-block my-4">Connect with Cal-ITP</h3>
...
       <p class="text-white text-start px-lg-4">
         See our latest milestones on our
         <a class="text-white fw-bold" href="/press">Press page</a>, and subscribe to the
-        <a class="text-white fw-bold"
-           href="https://lp.constantcontactpages.com/su/eLbtFoE/calitp?website"
-           rel="noreferrer"
-           target="_blank">Caltrans Mobility Newsletter</a>, a free biweekly resource with frequent Cal-ITP project updates.
+        <a
+          class="text-white fw-bold"
+          href="https://lp.constantcontactpages.com/su/eLbtFoE/calitp?website"
```
cmajel, line 62: This is also outside of this PR, but noting this newsletter link is no longer active (on prod or 11ty).

cmajel:
Alright added the items I saw. I can create follow on issues for the items I noted that may be outside of this PR.

src/index.liquid
cmajel: Follow on here: https://github.com/cal-itp/calitp.org/issues/628

src/_includes/contact.html
cmajel: Follow in issue here: https://github.com/cal-itp/calitp.org/issues/630

jgravois:
great catches @cmajel! we will patch both of these issues up and let you know when we're ready for you to take another look. 🙏

jgravois (approved):
i approve (w/ one hand admittedly in the cookie jar). 🍪

cmajel:
Yay the patches look fantastic! 🎉 

I discovered another small thing. I noticed when testing link previews there seems to be different logic and some line breaks on the 11ty preview descriptions for press releases.

I would expect it to be pulling a description from the first paragraph, which 11ty seems to be doing. But it is a change so I want to confirm that it was intentional.

Re linebreaks this seems to happen some times, but not all the time, and from what I saw it seemed to be isolated to press releases. 

I'm comfortable with this being a fast follow if it's not a quick fix.

A few examples. 

Comparisons:
* https://11ty--cal-itp-website.netlify.app/press/cal-itp-partners-celebrate-major-milestone-implementation-ods/
* https://www.calitp.org/press/cal-itp-partners-celebrate-major-milestone-implementation-ods
Screenshot 2026-05-04 at 7 00 35 PM

* https://11ty--cal-itp-website.netlify.app/press/cal-itp-benefits-launch/
* https://www.calitp.org/press/cal-itp-benefits-launch
Screenshot 2026-05-04 at 7 00 18 PM

jgravois:
this is another good catch @cmajel.

the new descriptions come from front-matter leads that were hand-written for these press pages.

https://github.com/cal-itp/calitp.org/blob/11ty/src/press/cal-itp-benefits-launch.md?plain=1#L8

right now on production these leads are ignored even though there is logic in place to extract them because jekyll's seo plugin does what you expected and just grabs the first content on the page it sees. (see https://github.com/cal-itp/calitp.org/pull/618#discussion_r3157840138 for more info)

I hadn't noticed until you documented it, but the line breaks in the yaml frontmatter are indeed preserved in the output shareable snippet.

we could investigate:
* collapsing that text into a single line (easy fix)
* using a different linebreak character in the markdown (might be easy too)
* ditching the leads altogether (if you think they're more trouble than they're worth)

but I agree with treating 👆 as a followup.

if you can you give this PR a genuine approval whenever you're sitting at your laptop next, we'll release and tee up the first set of fast-follows.

thanks so much for the thorough review!

cmajel (approved):

cmajel:
Sweet and approved!

>we could investigate:
> * collapsing that text into a single line (easy fix)
> * using a different linebreak character in the markdown (might be easy too)
> * ditching the leads altogether (if you think they're more trouble than they're worth)

My preference would be to go for option 1. And re leads my instinct is if there are prewritten leads for sharing, let's use them and where we don't, pulling the first sentence or two from the first paragraph feels like a good fallback. The leads are useful for social engagement so let's keep them!

jgravois:
@angela-tran merge at will :shipit: 

afterward i can confirm that google analytics is indeed still wired up correctly.

angela-tran:
Yep @jgravois, I agree that we should merge this soon. I wanted to make sure we didn't lose @cmajel's feedback about the link preview descriptions since it seems that is separate from the linebreaks. I'll create a ticket for it and then merge this!

**update**: this is the feedback I wanted to make sure we address

> And re leads my instinct is if there are prewritten leads for sharing, let's use them and where we don't, pulling the first sentence or two from the first paragraph feels like a good fallback.

I went through all the press releases and found that all the non-external ones will have some sort of specific preview description through either their `lead` or `intro` value. Basically I wanted to see if there would be any non-external press releases that would end up with the generic site description as their link preview description, which would then need to be addressed. Luckily there will not be, so nothing more to do here / no ticket needed!

PR #592, opened 2026-04-03
Chore: passing 11ty build

angela-tran:
Closes #585 

Worked with @jgravois to get these minimal changes that allow the 11ty build to succeed. Running `npm run build` results in a generated site.


jgravois (approved):
the deploy error is _Jekyll_ failing to build the site, which is to be expected at this point.

thanks for detangling @angela-tran!

Issue #591, opened 2026-04-03
chore: remove rvm, gemfile, yml etc.


jgravois:
once we're ready to cutover and have 11ty host the live site, the last cleanup step is to remove our Jekyll dependency and configuration files

* update README
* stop installing rvm in .devcontainer/Dockerfile
* remove Gemfile/Gemfile.lock
* remove .ruby-version file
* update mise.toml file
* remove src/_data/*.yml files
* remove outdated build command in Netlify UX

PR #593, opened 2026-04-03
Chore: Netlify config

angela-tran:
Closes #586 

These changes make it so Netlify is able to run 11ty and generate the site.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -3,4 +3,4 @@ node = "24"
 
 [settings]
 # https://github.com/jdx/mise/discussions/4345
-idiomatic_version_file_enable_tools = ["ruby"]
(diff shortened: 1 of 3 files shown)


mise.toml
```diff
@@ -3,4 +3,4 @@ node = "24"
 
 [settings]
 # https://github.com/jdx/mise/discussions/4345
-idiomatic_version_file_enable_tools = ["ruby"]
+idiomatic_version_file_enable_tools = ["node"]
```
jgravois, line 6: since we're not using a `.node-version` file in this repo, when we tackle #591 we _might_ be able to get rid of this declaration altogether. 🤙

jgravois (approved):
🚀

PR #594, opened 2026-04-03
Chore: minor cleanup

angela-tran:
Part of #587 

These are minor clean-up changes that would be good to get in before @jgravois and I split up the work for #587.


jgravois (approved):
planting the flag! 🏴‍☠️

PR #596, opened 2026-04-03
chore: update liquid-aware autoformatter

jgravois:
this PR is a reboot of https://github.com/cal-itp/calitp.org/pull/537 (and a bit of an improvement over https://github.com/cal-itp/mobility-marketplace/pull/824)

on the `11ty` branch right now, pre-commit and VS Code can't agree on how to format liquid code. this PR proposes that we drop the jinja flavored djlint pre-commit step in favor of prettier itself (w/ the shopify plugin enabled).

going forward, to get the formatting we want in VS Code _and_ in pre-commit, we're going to need to update the extensions on more files from `.html` to `.liquid`. it seems ideal to just do it as we go instead of injecting another noisy step 0 commit, but i'm happy to discuss alternatives.

🚧 ~reverting to draft to fix the CI error.~ 🚧

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,12 +31,11 @@ repos:
       - id: check-added-large-files
         args: ["--maxkb=2500"]
 
-  - repo: https://github.com/djlint/djLint
-    rev: v1.36.4
+  - repo: https://github.com/JoC0de/pre-commit-prettier
+    rev: v3.8.1
     hooks:
-      - id: djlint-reformat-jinja
-        files: \.html$
-        types_or: ["html"]
+      - id: prettier
+        types_or: [css, html, liquid, markdown, yaml]
 
   - repo: https://github.com/streetsidesoftware/cspell-cli
     rev: v9.6.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,11 @@ repos:
     rev: v3.8.1
     hooks:
       - id: prettier
+        additional_dependencies:
+          - prettier@3.8.1
+          - "@shopify/prettier-plugin-liquid@1.10.2"
         types_or: [css, html, liquid, markdown, yaml]
+        stages: [commit-msg]
 
   - repo: https://github.com/streetsidesoftware/cspell-cli
     rev: v9.6.0
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,12 +9,5 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
-  "editor.tabSize": 2,
-  "files.associations": {
-    "*.html": "liquid"
-  },
-  "[liquid]": {
-    "editor.defaultFormatter": "sissel.shopify-liquid",
-    "editor.formatOnSave": true
-  }
+  "editor.tabSize": 2
 }
--- a/src/_includes/footer.liquid
+++ b/src/_includes/footer.liquid
@@ -6,22 +6,19 @@
           <a class="text-decoration-none text-white" rel="noreferrer" href="#">Back to top</a>
         </li>
         <li class="col-md-auto col-6 mb-1 mb-md-0">
-          <a class="text-decoration-none text-white"
-             rel="noreferrer"
-             target="_blank"
-             href="https://dot.ca.gov/privacy-policy">Privacy policy</a>
+          <a class="text-decoration-none text-white" rel="noreferrer" target="_blank" href="https://dot.ca.gov/privacy-policy"
+            >Privacy policy</a
+          >
         </li>
         <li class="col-md-auto col-12 my-1 my-md-0">
-          <a class="text-decoration-none text-white"
-             rel="noreferrer"
-             target="_blank"
-             href="https://www.ca.gov/use/">Conditions of use</a>
+          <a class="text-decoration-none text-white" rel="noreferrer" target="_blank" href="https://www.ca.gov/use/"
+            >Conditions of use</a
+          >
         </li>
         <li class="col-md-auto col-12 my-1 my-md-0">
-          <a class="text-decoration-none text-white"
-             rel="noreferrer"
-             target="_blank"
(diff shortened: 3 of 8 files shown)


src/_includes/footer.liquid
jgravois, line 1: this file is just an example of the autoformatting in VS Code. i tried manually saving a deviation and confirmed that pre-commit wouldn't accept it.

.vscode/settings.json
```diff
@@ -9,12 +9,5 @@
...
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
-  "editor.tabSize": 2,
-  "files.associations": {
-    "*.html": "liquid"
-  },
-  "[liquid]": {
-    "editor.defaultFormatter": "sissel.shopify-liquid",
-    "editor.formatOnSave": true
-  }
```
jgravois, line 19: its nice that VS Code can treat plain .html files as liquid, but since pre-commit cannot, we're better sticking to the defaults.

src/_layouts/default.html
```diff
@@ -1,22 +1,18 @@
...
-
-    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/js/bootstrap.bundle.min.js"
-            integrity="sha384-YvpcrYf0tY3lHB60NNkmXc5s9fDVZLESaAA55NDzOxhy9GkcIdslK1eN7N6jIeHz"
-            crossorigin="anonymous"></script>
-  </body>
-</html>
+<!doctype html>
+<html lang="en">
+  <head>
+    {% include "meta.html" %} {% include "styles.html" %} {% include "analytics.html" %}
```
jgravois, line 4: the formatting here will be improved when we give this file a .liquid extension too. 😇

.pre-commit-config.yaml
```diff
@@ -31,12 +31,11 @@ repos:
       - id: check-added-large-files
         args: ["--maxkb=2500"]
 
-  - repo: https://github.com/djlint/djLint
-    rev: v1.36.4
+  - repo: https://github.com/JoC0de/pre-commit-prettier
+    rev: v3.8.1
```
jgravois, line 35: unlike in benefits, since this is a Node.js project now, i'm proposing that we install the version of prettier that we want explicitly instead of relying on what comes bundled in the VS Code extension.

.pre-commit-config.yaml
```diff
@@ -31,12 +31,15 @@ repos:
...
     hooks:
-      - id: djlint-reformat-jinja
-        files: \.html$
-        types_or: ["html"]
+      - id: prettier
+        additional_dependencies:
+          - prettier@3.8.1
+          - "@shopify/prettier-plugin-liquid@1.10.2"
+        types_or: [css, html, liquid, markdown, yaml]
+        stages: [commit-msg]
```
jgravois, line 42: it would be ideal to run prettier on _both_ pre-commit stages, but I'd like to save solving that riddle for another day.

angela-tran:
Reviewing this now :eyes: 

I'm aiming to not dwell too much on minute details so we can get past this, but I do want to spend some time making sure I understand the discussion from #537 and the overall picture of our autoformatting setup across MobiMart and calitp.org.

angela-tran:
I wrote this summary out as a way of wrapping my head around the current status of our autoformatting for our static sites and thought I'd go ahead and post it in case it helps anyone else.

In #537, we got pretty far with making calitp.org use `prettier` + `@shopify/prettier-plugin-liquid` for pre-commit and the Shopify Liquid extension for when we're editing files in VS Code.

There were 2 blockers, as noted in that PR description:
- When running `prettier` either through command-line or the pre-commit hook, the result did not match the result of running the Shopify Liquid VS Code extension.
- Jekyll would fail to build the code formatted with the Shopify Liquid VS Code extension.

In https://github.com/cal-itp/mobility-marketplace/pull/824, the project by that point had been converted over to 11ty, so the second blocker was no longer an issue, and we were all clear to add the Shopify Liquid VS Code extension. The pre-commit check was not added because the first blocker was still an issue. I'm guessing / gathering that the re-formatting that was done in 23ca85d21c (and recorded in [`.git-blame-ignore-revs`](https://github.com/cal-itp/mobility-marketplace/pull/824/changes#diff-f247fbfbe928b907db42554d0b3006b28dd11c25a59be031abda73b11a2c934aR2)) was done with the VS Code extension.

In this PR, we learned that the `@shopify/prettier-plugin-liquid` plugin will only run on `.liquid` files, so we now have a path forward for resolving the first blocker: ultimately, all our templates need to be re-saved from `.html` to `.liquid` files.

angela-tran (approved):
I can see you've done a lot of untangling here, and it all makes sense to me. I agree it's probably best to go ahead and get this in and update the files as we touch them. 🚀

jgravois:
> I'm guessing / gathering that the re-formatting ... was done with the VS Code extension.

thats exactly right.

> In this PR, we learned that the @shopify/prettier-plugin-liquid plugin will only run on .liquid files

this is correct too. to be even more specific, when the plugin is available and configured correctly, we can rely on prettier itself to auto-format correctly in VS Code, in the CLI and from pre-commit.

i'll create a follow-up MobiMart issue to re-enable pre-commit, but i think we should land a few more PRs here first to confirm that the ergonomics are what we want/need.

angela-tran:
> ... to be even more specific, when the plugin is available and configured correctly, we can rely on prettier itself to auto-format correctly in VS Code, in the CLI and from pre-commit.

Oh interesting. It would be nice to simplify it all down to one tool. ~It sounds then like we might move towards removing~

https://github.com/cal-itp/calitp.org/blob/f6b5b71147caaca621d438a564f79d780f6582e0/.devcontainer/devcontainer.json#L26

~and `"shopify.theme-check-vscode"` for [MobiMart](https://github.com/cal-itp/mobility-marketplace/pull/824/changes#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R26).~

~Anyways, yep we can look into these auto-formatting follow-ups later.~

The VS Code extension provides features aside from autoformatting, so there's still value in having it; it'll just be that we let `prettier` do all the autoformatting.

PR #597, opened 2026-04-03
chore: get 11ty loading stylesheets

jgravois:
Part of #587

just a few small tweaks to fix the busted CSS.

the crux of the issue is that 11ty didn't recognize src/stylesheets/dsdl/tokens.css as a liquid template automatically.

Screenshot 2026-04-03 at 2 41 34 PM

### DIFF
--- a/.cspell/dev-dictionary.txt
+++ b/.cspell/dev-dictionary.txt
@@ -77,6 +77,8 @@ smallcaps
 smartsymbols
 stylesheet
 superfences
+tbody
+thead
 toc
 tombi
 twemoji
--- a/src/dsdl.liquid
+++ b/src/dsdl.liquid
@@ -3,12 +3,8 @@ title: DSDL demo
 layout: default
 stylesheet: dsdl-demo.css
 ---
-
-{% comment %} djlint:off {% endcomment %}
-
 <div class="dsdl-demo">
-
-  {% include "clipped-header.html" %}
+  {% include 'clipped-header.html' %}
 
   <div class="row justify-content-center">
     <div class="clipped-header background-calitp-blue col-11 col-md-10 mt-4 mt-md-5 mb-4 py-4 py-md-5 px-4 px-md-0">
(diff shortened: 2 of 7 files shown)


.cspell/dev-dictionary.txt
```diff
@@ -77,6 +77,8 @@ smallcaps
 smartsymbols
 stylesheet
 superfences
+tbody
+thead
```
jgravois, line 81: there's probably a smarter way to ensure that all valid html tags are recognized in .liquid files, but this seems harmless enough.

src/dsdl.liquid
angela-tran, line 1: Seeing all the formatting-only changes in this file, I wonder if we should start a `.git-blame-ignore-revs` file and split out commits for formatting-only changes. What do you think @jgravois? We don't have to go back and do it for this one; it's just an idea for subsequent PRs.

jgravois: now that I have confidence that we can configure the prettier CLI to autoformat files consistently, my instinct is that our time would be better spent updating the extension of any/all files that aren't formatted well as .html and then doing/ignoring one bulk autoformat commit for the entire repo.

if you like the idea of doing it file by file better, thats okay with me too though.

jgravois:
i was expecting netlify to autodeploy a preview branch when the base branch of this PR became `11ty`, but it did not.

i'm going to rebase and force push without making any other changes to see if i can kickstart it.

src/stylesheets/dsdl/tokens.css.liquid
```diff
@@ -48,10 +48,12 @@
 /* COLORS */
 
 :root {
-  {% for color in site.data.dsdl_colors -%}
+  {% for color in dsdlColors -%}
   --dsdl-{{ color.group }}-{{ color.level }}: {{ color.hex }};
   {% endfor %}
+}
 
+:root {
```
angela-tran, line 56: Does this change anything?

jgravois: sorry, I forgot to call out specifically. it doesn't change anything at all in the live site. it's purely a way to minimize the number of erroneous linting errors displayed in VS Code.

you can see there is already another `:root` CSS declaration immediately below in this file on `main`.

see https://github.com/cal-itp/calitp.org/pull/537#issuecomment-3648257406 for more info.

src/dsdl.liquid
angela-tran: Ok, yeah let's go for the one bulk autoformat

src/stylesheets/dsdl/tokens.css.liquid
angela-tran: Ah I see, thanks!

angela-tran (approved):

PR #599, opened 2026-04-06
chore: reintroduce navigation links in the header

jgravois:
Part of #587


.cspell/dev-dictionary.txt
```diff
@@ -77,8 +77,6 @@ smallcaps
 smartsymbols
 stylesheet
 superfences
-tbody
-thead
```
jgravois, line 81: the second commit in this PR ensures that _all_ valid html tags are recognized in .liquid templates.

src/_data/site.js
jgravois, line 1: i accidentally committed this file in the root of the repo instead of the _data/ directory initially.

src/_includes/header.liquid
jgravois, line 1: imo, the auto-formatting here is a _lot_ more legible. 🛁

angela-tran (approved):

PR #600, opened 2026-04-06
chore: declare more explicit liquid templates and let prettier autoformat files

jgravois:
as discussed in https://github.com/cal-itp/calitp.org/pull/597#discussion_r3041431226

the first commit in this PR adds a few helpful formatting npm scripts, renames more .liquid templates and manually fixes a few reported prettier errors.

the second commit is the result of running `npm run lint:fix`

the third ensures that the prior commit is ignored by `git blame`.


src/index.liquid
```diff
@@ -20,7 +20,7 @@ <h1 class="text-center text-lg-start pt-5">A modern and consistent transportatio
     <picture>
       <img class="hero-header w-100 mx-auto mt-lg-5 pt-lg-4 pb-lg-0 py-3 pb-4"
            src="images/hero-header.png"
-           alt="A trio of images, clockwise from top: a bus, a train platform with a sign that announces “Next train in 3 minutes,” and a transit rider paying their fare by tapping their smartphone’s mobile wallet on a payment reader when boarding" />
+           alt="A trio of images, clockwise from top: a bus, a train platform with a sign that announces &quot;Next train in 3 minutes,&quot; and a transit rider paying their fare by tapping their smartphone’s mobile wallet on a payment reader when boarding" />
```
jgravois: prettier expects that the curved quotes here should close the entire html tag. since its an alt tag, i switched to encoded quotes, but happy to discuss alternatives.

src/press.html
```diff
@@ -16,9 +16,9 @@ <h1 class="text-white mb-2">Press</h1>
       Below you’ll find news about Cal-ITP and our initiatives, including press releases and media coverage about new launches and project milestones. Interested in getting in touch? Reach out to us at <a rel="noreferrer"
     target="_blank"
     class="fw-bold text-white"
-    href="mailto:hello@calitp.org">hello@calitp.org</a>.</span>
```
jgravois, line 19: this is an invalid closing `<span>` tag that prettier caught out.

src/_includes/script_redirect.liquid
```diff
@@ -1,3 +1,3 @@
 <h1>Redirecting...</h1>
-<a href="{{ include.location }}">Click here if you are not redirected.<a>
+<a href="{{ include.location }}">Click here if you are not redirected.</a>
```
jgravois, line 2: prettier caught out that we weren't closing this `<a>` tag correctly either.

angela-tran (commented):
Hey @jgravois - I'm seeing that Prettier and VS Code don't agree on how to format `.html` files and that the the pre-commit hook doesn't work how I'd expect. I think I know how to solve both these issues, so let's pair on this tomorrow to first make sure we're both seeing the same thing.

angela-tran (approved):
This PR gets us in a state where 
- everything except `docs/` and `mkdocs.yml` is formatted with Prettier
- we have a pre-commit hook that runs Prettier and works locally and with precommit.ci
- formatting done by VS Code is also using Prettier

PR #604, opened 2026-04-08
chore: reinstitute /press landing page and subpages

jgravois:
Part of #587

* https://deploy-preview-604--cal-itp-website.netlify.app/press/
* https://deploy-preview-604--cal-itp-website.netlify.app/press/cal-itp-expands-contactless-payment-options/
* etc.


src/_includes/article.liquid
jgravois, line 1: the crux of the difference here is that custom front-matter properties in 11ty are nested within a `data` prop when you loop through the items in a collection

src/_includes/article.liquid
```diff
@@ -10,24 +10,27 @@
...
+    {{- item.data.title -}}
   </a>
   <br>
   <span class="text-body-tertiary text-body-m">
-    {% if item.outlet %}{{ item.outlet }} |{% endif %}
-    {% if item.tags.size > 0 %}{{ item.tags | join: ', ' }} |{% endif %}
+    {% if item.data.outlet %}{{ item.data.outlet }} |{% endif %}
+    {% if item.data.tags.size > 0 %}
+      {% assign filtered_tags = item.data.tags | where_exp: 't', "t != 'press'" %}
+      {{ filtered_tags | join: ', ' }} |
```
jgravois, line 32: the change here is a function of the data cascade merging the 'press' tag for each item in the collection with the custom tags declared in front-matter.

src/_layouts/press.liquid
```diff
@@ -1,14 +1,14 @@
...
+{% if external %}
+  {% include 'script_redirect', location: external %}
 {% endif %}
 
 {% capture date %}
-  {% include "date" date: page.date format: "%B %e, %Y, " %}
+  {% include "date" date: date format: "%B %e, %Y, " %}
 {% endcapture %}
-{% assign location = page.location | default: 'Sacramento, CA' %}
+{% assign location = location | default: 'Sacramento, CA' %}
```
jgravois, line 11: outside of a collection loop, custom front-matter properties don't hang off the 'page' in 11ty, they are their own global variables.

src/press/cal-itp-announces-ods.md
```diff
@@ -76,4 +76,4 @@ for a full list of members.
 
 ## Questions? In need of free support?
 
-For additional information, please email [hello@calitp.org](mailto:hello@calitp.org?subject=Attn: Scott Frazier) Attn: Scott Frazier.
+For additional information, please email [hello@calitp.org](mailto:hello@calitp.org?subject=Attn:%20Scott%20Frazier) Attn: Scott Frazier.
```
jgravois, line 79: i noticed that 11ty's markdown parser short-circuits when the url has a space in it. 🤷

src/press/press.json
jgravois, line 1: this PR was a good excuse for me to finally wrap my mind around the [data cascade](https://www.11ty.dev/docs/data-cascade/) that @angela-tran noticed in the 11ty docs back in the day.

the gist of it is that creating a .json file with the same name as the parent directory is a convenient mechanism to merge common properties for all sibling pages with custom front-matter declared in individual files.

in practice this means that the 'press' tag declared here that defines our collection ends up merged with the tags defined in front-matter.

```
tags: ["press", "GTFS"]
```
ref: https://www.11ty.dev/docs/data-template-dir/#examples

angela-tran (commented):
The generated /press page looks great. A few notes for cleaning this up:
- Based on commit messages, it looks like changes meant for your third commit ended up in your second commit.
- I'm not seeing a clear pattern for when we use single-quotes vs. double-quotes. What is causing this inconsistency? Can we standardize this?
- The changes that are not related to getting the /press page working -- let's separate them out into their own PRs. I know with all this 11ty conversion work that we might not be able to keep every PR narrowly scoped and atomic, but we should still try to stick to that practice as much as possible.

jgravois:
> The changes that are not related to getting the /press page working -- let's separate them out into their own PRs.

works for me. 🤙 

> I'm not seeing a clear pattern for when we use single-quotes vs. double-quotes. What is causing this inconsistency? Can we standardize this?

good question.
* prettier itself defaults to [`singleQuote: false`](https://prettier.io/docs/options#quotes)
* the liquid plugin defaults to [`liquidSingleQuote: true`](https://shopify.dev/docs/storefronts/themes/tools/liquid-prettier-plugin) 

personally i don't have many/any opinions about ☝️. i imagine there are good reasons the defaults diverge but i'm happy to explore changing either of these in `.prettierrc.js` and evaluating the CLI output in a separate PR.

angela-tran:
> > I'm not seeing a clear pattern for when we use single-quotes vs. double-quotes. What is causing this inconsistency? Can we standardize this?
> 
> good question.
> 
>     * prettier itself defaults to [`singleQuote: false`](https://prettier.io/docs/options#quotes)
> 
>     * the liquid plugin defaults to [`liquidSingleQuote: true`](https://shopify.dev/docs/storefronts/themes/tools/liquid-prettier-plugin)
> 
> 
> personally i don't have many/any opinions about ☝️. i imagine there are good reasons the defaults diverge but i'm happy to explore changing either of these in `.prettierrc.js` and evaluating the CLI output in a separate PR.

Yeah, I see the inconsistency was introduced in 029b9c5585f81e79e68100091979e8e5a390ffea and slipped past our review. I'm ok with just continuing on with what we have configured right now, and at the end, we can take a final pass through the formatting and possibly do another bulk re-formatting.

src/press/press.json
```diff
@@ -0,0 +1,5 @@
+{
+  "tags": ["press"],
```
angela-tran, line 2: So just to restate some of https://github.com/cal-itp/calitp.org/pull/604/changes#r3053781804, this makes it so all files under `/press` have `press` in their `tags`  and therefore be a part of `collections.press`. But then we also have our "`tags`" (`GTFS`, `Contactless Payments`, `Benefits`, `Data Plans`) for organizing our content for filtering purposes.

src/press.liquid
```diff
@@ -41,8 +41,8 @@ description: >
         >
           {% include 'articles', items: items %}
         </div>
-        {% for tag in site.data.press_tags %}
-          {% assign items = site.press | sort: 'date' | reverse | where_exp: 'item', 'item.tags contains tag.name' %}
+        {% for tag in pressTags %}
+          {% assign items = collections.press | sort: 'date' | reverse | where_exp: 'item', 'item.data.tags contains tag.name' %}
```
angela-tran, line 45: Just leaving behind breadcrumbs as we encounter more 11ty / Jekyll differences

https://www.11ty.dev/docs/collections/

vs.

https://jekyllrb.com/docs/collections/

angela-tran (approved):
Looks good!

src/press/press.json
jgravois: exactly 🎯

PR #606, opened 2026-04-09
chore: misc 11ty fixes

jgravois:
just a handful of small/one-off changes that i've been trying to sneak in via arbitrary PRs. 😈

### DIFF
--- a/.devcontainer/compose.yml
+++ b/.devcontainer/compose.yml
@@ -7,6 +7,8 @@ services:
       dockerfile: .devcontainer/Dockerfile
     image: calitp-dot-org:site
     command: sleep infinity
+    ports:
+      - 8080:8080
     volumes:
       - ..:/usr/src/calitp
 
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,10 +1,14 @@
+// if the plugin was ESM compatible, we could use await import('@shopify/prettier-plugin-liquid') instead of shimming in require
+// https://stackoverflow.com/questions/69099763/referenceerror-require-is-not-defined-in-es-module-scope-you-can-use-import-in
+import { createRequire } from "module";
+const require = createRequire(import.meta.url);
+
 // Allow Prettier in .pre-commit-config.yaml to find plugins.
 // https://stackoverflow.com/questions/78708497/prettier-plugins-not-found-with-pre-commit
 // https://github.com/prettier/prettier/issues/15085
-// we can explore making this an ESM module sometime down the line
-module.exports = {
-  tabWidth: 2,
-  useTabs: false,
-  printWidth: 130,
+const config = {
   plugins: [require.resolve("@shopify/prettier-plugin-liquid")],
+  printWidth: 130,
 };
+
+export default config;
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,4 +1,6 @@
-// if the plugin was ESM compatible, we could use await import('@shopify/prettier-plugin-liquid') instead of shimming in require
+// this convoluted plugins syntax is needed to ensure availability in the pre-commit context specifically
+// if the shopify plugin was ESM compatible, we could use await import() instead
+// we also tried swapping in import.meta.url(). that works fine locally, but causes inexplicable behavior on pre-commit.ci
 // https://stackoverflow.com/questions/69099763/referenceerror-require-is-not-defined-in-es-module-scope-you-can-use-import-in
 import { createRequire } from "module";
 const require = createRequire(import.meta.url);
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,6 +3,3 @@ docs/
 
 # this file has custom formatting we don't want prettier to revert
 src/stylesheets/dsdl/tokens.css.liquid
-
-# TODO: figure out how to uncomment this safely
-mkdocs.yml
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,9 +4,26 @@
   "version": "2.0.0",
   "tasks": [
     {
-      "label": "Jekyll: Build and Serve Locally",
+      "label": "11ty: Serve",
       "type": "shell",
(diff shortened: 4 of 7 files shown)


src/_data/site.js
jgravois, line 1: in the `11ty` branch right now `npm start` prints the warning below 👇

```
(node:18950) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///usr/src/calitp/eleventy.config.js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /usr/src/calitp/package.json.
```
this is fixed by e08e7309b1d4a68f1ba06611fb7f3b32730e4a86

.prettierrc.js
```diff
@@ -1,10 +1,14 @@
...
+import { createRequire } from "module";
+const require = createRequire(import.meta.url);
+
 // Allow Prettier in .pre-commit-config.yaml to find plugins.
 // https://stackoverflow.com/questions/78708497/prettier-plugins-not-found-with-pre-commit
 // https://github.com/prettier/prettier/issues/15085
-// we can explore making this an ESM module sometime down the line
-module.exports = {
-  tabWidth: 2,
-  useTabs: false,
```
jgravois, line 7: these two values weren't customizing anything. they are prettier's defaults.

https://prettier.io/docs/options#tab-width

.prettierrc.js
```diff
@@ -1,10 +1,14 @@
+// if the plugin was ESM compatible, we could use await import('@shopify/prettier-plugin-liquid') instead of shimming in require
+// https://stackoverflow.com/questions/69099763/referenceerror-require-is-not-defined-in-es-module-scope-you-can-use-import-in
+import { createRequire } from "module";
+const require = createRequire(import.meta.url);
```
angela-tran: I spent a little time refreshing my memory on CommonJS vs. ESM

CommonJS uses `require`
ESM uses `import` / `export`

Rather than shimming in `require` like this, it looks like we can just change the code on line 10 to use `import` instead.

```diff
- plugins: [require.resolve("@shopify/prettier-plugin-liquid")],
+ plugins: [import.meta.resolve("@shopify/prettier-plugin-liquid")],
```

That issue you linked to on line 8 was helpful -- that's where I saw that we could use `import.meta` (https://github.com/prettier/prettier/issues/15085#issuecomment-1636993346).

You have more JS experience than me, so you're probably a better judge than me on if this is better or not. Let me know what you think

.devcontainer/compose.yml
```diff
@@ -7,6 +7,8 @@ services:
       dockerfile: .devcontainer/Dockerfile
     image: calitp-dot-org:site
     command: sleep infinity
+    ports:
+      - 8080:8080
     volumes:
```
angela-tran, line 12: I generally prefer to avoid adding things without fully understanding why they're necessary, and things _seem_ to work for me without this. I can't tell for sure though, and I'm going crazy with repeatedly rebuliding the devcontainer to see if I can get it to break, so I'm just going to accept this 😅 

I guess we at least have this comment to refer back on if there's any question on why we added this.

angela-tran (changes requested):
1 suggested change on the plugin import

.prettierrc.js
jgravois: i glossed over something a few days ago indicating that the `file://` url protocol output by `import.meta.resolve` wasn't 100% compatible and didn't even bother to test it.

but now that i have, i agree that what you're suggesting is an improvement! 58674d7

.devcontainer/compose.yml
jgravois: > things seem to work for me without this

sorry, based on what i saw in [benefits](https://github.com/cal-itp/benefits/blob/7657f9e0c7c5f32491fd3a0a9cf9275781db6f44/compose.yml#L27) i thought this was common practice, but don't be fooled. i haven't fully wrapped my mind around it either. 😇 

_sometimes_ when i rebuild the container i end up with 11ty serving the app on `8080`, but not accessible via that port in the browser. my working theory is that docker takes the liberty of forwarding that port to another arbitrary port when/if needed.

when this occurs, in most of our other projects i have been able to inspect the ports tab in VS Code and find that mapping, but in this project when that happens it has been a mystery.

Screenshot 2026-04-11 at 6 37 45 AM

the syntax i'm suggesting is purely an attempt to force docker to bind `8080` explicitly on the local machine to the port used by 11ty inside the container.

all that said, i'd still like to learn more about:
* why/when arbitrary remapping occurs
* what controls whether or not the forwarding is advertised in VS Code

if you'd rather i pursue that before we merge this change, that's not a problem for me. the temporary workaround is just to define user-specific port forwarding for `8080` in VS Code within the container session.

.prettierrc.js
jgravois: hmm. your suggestion worked just fine for me locally, but it caused an unexpected follow-up commit (50f05e3) when i pushed 58674d7.

if i'm reading that diff correctly, the liquid plugin is working but our custom `printWidth` is ignored. if the node runtime didn't support `import` i would have expected the opposite.

i'll play around a little and see what i can figure out. worst-case scenario i guess we can just revert back to what i suggested initially. 🤷

jgravois: no joy, so i just reverted.

you can see a couple things i tried to get more explicit control over the node runtime in the pre-commit ci environment in https://github.com/cal-itp/calitp.org/compare/chore/misc-experiment?expand=1

angela-tran: That's a bummer, thanks for trying it and looking into all that :pray:

.devcontainer/compose.yml
angela-tran: Thanks for this explanation. Having this context written out makes me feel better about punting on this for now ☺️

angela-tran (approved):

PR #608, opened 2026-04-10
Chore: reinstitute Resources page and subpages

angela-tran:
Part of #587 

https://deploy-preview-608--cal-itp-website.netlify.app/resources
https://deploy-preview-608--cal-itp-website.netlify.app/resources/fact-sheet-merchant-category-codes-in-mobility/


angela-tran:
~I noticed both the `/resources` and `/press` pages are missing their page titles, so that's a thing that still needs to be fixed.~ Eh, maybe we make that its own ticket to fix the page title across the whole site

angela-tran:
Ready for review - leaving the page title fix to be done as a separate PR

src/_includes/article.liquid
```diff
@@ -28,7 +28,7 @@
   <span class="text-body-tertiary text-body-m">
     {% if item.data.outlet %}{{ item.data.outlet }} |{% endif %}
     {% if item.data.tags.size > 0 %}
-      {% assign filtered_tags = item.data.tags | where_exp: 't', "t != 'press'" %}
+      {% assign filtered_tags = item.data.tags | where_exp: 't', "t != 'press' and t != 'resources'" %}
```
jgravois: i know that i was the one to introduce this pattern, but after i saw your great tweak to the eleventy.config.js file i remembered we could separate concerns a lot more nicely by leveraging a [computed property](https://www.11ty.dev/docs/data-computed/) for this.

```js
// src/resources/resources.11tydata.js
export default {
  eleventyComputed: {
    // omit the collection tag for purposes of client-side filtering
    filteredTags: (data) => data.tags.filter((t) => t !== "resources"),
  },
};
```

```liquid
{% if item.data.filteredTags.size > 0 %}
  {{ item.data.filteredTags | join: ', ' }} |
{% endif %}
```

eleventy.config.js
```diff
@@ -15,4 +15,11 @@ export default async function (eleventyConfig) {
   eleventyConfig.addPassthroughCopy("**/*.svg");
 
   eleventyConfig.addPassthroughCopy("src/assets/");
+
+  // Define a custom collection for grouping resources by category.
+  eleventyConfig.addCollection("resourcesByCategory", (collectionApi) => {
+    const resources = collectionApi.getFilteredByTag("resources");
+    const groupedByCategory = Object.groupBy(resources, (resource) => resource.data.category);
```
jgravois, line 23: i'd never seen [`Object.groupBy()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) before. very cool!

(especially since we don't have to worry about clients supporting it)

eleventy.config.js
```diff
@@ -15,4 +15,11 @@ export default async function (eleventyConfig) {
...
+    return Object.entries(groupedByCategory).sort((a, b) => a[0].localeCompare(b[0])); // order by category name in ascending order
```
jgravois: i'm 99% sure that a plain ol `.sort()` would work just as well here.
```suggestion
    return Object.entries(groupedByCategory).sort(); // order by category name in ascending order
```

jgravois (changes requested):
overall this is looking really slick!

edge-case: i noticed when a resource has no category we're displaying the pipe all by its lonesome.

Image

i think my suggestion to use a computed property will fix that. 😈

src/_data/resourceTags.json
jgravois, line 1: i just realized this is a subset of `pressTags`. we could DRY things up like this:

```js
// _data/pressTags.js
import resourceTags from "./resourceTags.json" with { type: "json" };

export default function () {
  return [{ name: "Data Plans", id: "data-plans" }, ...resourceTags];
}
```

src/resources/resources.json
jgravois, line 1: definitely not a blocker in this PR, but i feel like using json/js to assign a category to resources in a sub-directory would be less brittle/duplicative than the way we're assigning strings in front-matter currently.

we can explore later this month if there's still time.

src/_includes/article.liquid
angela-tran: > edge-case: i noticed when a resource has no category we're displaying the pipe all by its lonesome.

Great catch and suggested fix. I saw the lonesome pipe at one point too and then forgot about it :woman_facepalming: 
Addressed this in [6bec2c4](https://github.com/cal-itp/calitp.org/pull/608/commits/6bec2c4d06729c2f474d2d5f61c541bb74d7798b) and [e04444c](https://github.com/cal-itp/calitp.org/pull/608/commits/e04444c9e440a6c0bbf58eed86f4372758213653)

eleventy.config.js
angela-tran: That does work, thanks! Changed in [85f383c](https://github.com/cal-itp/calitp.org/pull/608/commits/85f383c38cf295a2023f783dbf9c6316a1d7a0b7)

I meant to write a few notes here too about this refactor as a whole:

In Jekyll, the output of the [`group_by` filter](https://github.com/cal-itp/calitp.org/pull/608/changes#diff-9b134a38f8a28ef33a9717a1b4b323e96ce63c8d6c5bcd21790224cebc1eabc4L37) was an array like this:
```
[
  { "name": "category 1", "items": [...] },
  { "name": "category 2", "items": [...] }
]
```
So that's why we would refer to [`group.name` and `group.items`](https://github.com/cal-itp/calitp.org/pull/608/changes#diff-9b134a38f8a28ef33a9717a1b4b323e96ce63c8d6c5bcd21790224cebc1eabc4L47-L48) in our template.

In 11ty, there isn't an equivalent to the `group_by` filter, so we define this custom collection which uses `Object.groupBy` to get an object with the "groups" as properties, like this:
```
{
  "category 1": [...],
  "category 2": [...]
}
```

And then, so that we can sort the categories, we turn that object into an array of the key-value pairs using `Object.entries` and call `sort`:
```
[
  ["category 1", [...],
  ["category 2", [...]
]

src/_data/resourceTags.json
angela-tran: Changed this in [7ca022c](https://github.com/cal-itp/calitp.org/pull/608/commits/7ca022c4568ce132c9aec6fab9da2893c4496606) - note: "Data Plans" needs to appear last to preserve the order currently on the live site.

src/resources/resources.json
angela-tran: Yeah, it could be that there is a less duplicative way of doing that. If we decide to pursue that, we should also look into how the [Auto-Create PRs for Resource Requests](https://github.com/cal-itp/calitp.org/blob/main/.github/workflows/resource-template.yml#L38) would need to change.

angela-tran:
Ready for re-review @jgravois

jgravois (approved):
i'm lovin it! 🍔 

something is borked with the footnotes on this page, but lets patch that up in a follow-up PR.

https://deploy-preview-608--cal-itp-website.netlify.app/resources/fact-sheet-improving-financial-access-through-secured-credit-cards/

PR #609, opened 2026-04-14
fix: make footer sticky on /customer-stories

jgravois:
part of #587

right now its borked when you scroll to the bottom of [/customer-stories](https://11ty--cal-itp-website.netlify.app/customer-stories/)

Screenshot 2026-04-14 at 10 03 28 AM


angela-tran (approved):
☺️

PR #612, opened 2026-04-15
fix: get markdown footnotes rendering correctly

jgravois:
fix for the issue reported in https://github.com/cal-itp/calitp.org/pull/608#pullrequestreview-4109329326:

* gets the footnotes on [this page](https://deploy-preview-612--cal-itp-website.netlify.app/resources/fact-sheet-improving-financial-access-through-secured-credit-cards/) rendering correctly
* Doesn't introduce any other visual regressions in markdown-based pages

### DIFF
--- a/.cspell/dev-dictionary.txt
+++ b/.cspell/dev-dictionary.txt
@@ -47,6 +47,7 @@ jinja
 Kegan
 keyserver
 linenums
+linkify
 liquidjs
 Machiko
 Maher
--- a/eleventy.config.js
+++ b/eleventy.config.js
@@ -1,4 +1,13 @@
+import markdownIt from "markdown-it";
+import markdownItFootnote from "markdown-it-footnote";
+
 export default async function (eleventyConfig) {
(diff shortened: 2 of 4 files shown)


angela-tran (approved):
Nice!

I noticed this `markdown-it-footnote` plugin doesn't render a superscript number when there are multiple references to the same footnote, but this doesn't seem like something to worry about.

Jekyll-rendered footnotes:

Image

11ty-rendered footnotes:

Image

jgravois:
> I noticed this markdown-it-footnote plugin doesn't render a superscript number when there are multiple references to the same footnote

great catch! on the flip side, i do like the fact that the second instance of an inline reference gets its own unique suffix appended.

Screenshot 2026-04-16 at 8 58 54 AM

i'll try adding https://github.com/markdown-it/markdown-it-sup to the mix quickly to see if that fixes the problem you noticed. if not, i'll make a note of the disrepancy for when we ask for outside help for a final QA. i agree that it doesn't feel like a blocker.

edit: adding markdown-it-sup didn't help!

PR #614, opened 2026-04-24
chore: configure 11ty sitemap plugin

jgravois:
resolves #588.

* wires up an 11ty plugin to generate /sitemap.xml
* omits /404.html and other pages that we don't want to publicize
* fixes a few bugs related to outdated attempts to access custom front-matter props associated with individual pages that i caught out along the way. (CIL for more info)

### DIFF
--- a/.cspell/dev-dictionary.txt
+++ b/.cspell/dev-dictionary.txt
@@ -41,6 +41,7 @@ Grotesk
 gtag
 gtfs
 headerlink
+hubspot
 inlinehilite
 jekyll
 jinja
--- a/eleventy.config.js
+++ b/eleventy.config.js
@@ -1,7 +1,12 @@
+import sitemap from "@quasibit/eleventy-plugin-sitemap";
 import markdownIt from "markdown-it";
 import markdownItFootnote from "markdown-it-footnote";
 
+import site from "./src/_data/site.js";
+
 export default async function (eleventyConfig) {
+  eleventyConfig.addPlugin(sitemap, { sitemap: { hostname: site.url } });
+
   // this customization opts us into a markdown parsing feature that parses and
   // formats footnotes in a div of their own at the bottom of the page
   // Jekyll provides this functionality OOTB
--- a/src/404.html
+++ b/src/404.html
@@ -1,6 +1,7 @@
 ---
 layout: default
 permalink: /404.html
+eleventyExcludeFromCollections: true
 ---
 
 <article>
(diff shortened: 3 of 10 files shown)


angela-tran (approved):
This looks great!

I noticed a few differences between the currently live `sitemap.xml` and the one in this branch. I think these differences make sense to move forward with, but just to document it and/or make sure these differences are intentional:

The currently live `sitemap.xml` has 99 `<url>`s, the one in this branch has 61 `<url>`s. This is because:
- the items under `/assets` are no longer listed in the sitemap (there are 36 assets)
- we no longer list `/dsdl.html` 👍 
- we no longer list `/links.html` 🤔 🤷‍♀️ 

Screenshot from 2026-04-27 11-52-19

jgravois:
> the items under `/assets` are no longer listed in the sitemap (there are 36 assets)

great catch @angela-tran. my first instinct is that this is correct.

the .pdf files in `/assets` fall into two categories:

1. those referenced in front-matter and served up via a page that _is_ included in the sitemap. (like [this one](https://github.com/cal-itp/calitp.org/blob/4b8940d4f697ade4ce7a2081acb3012a30d47204/src/resources/case-study-holiday-service-transit-traveler-information-research.md))
1. those not referenced anywhere else in the HTML of the current site.

it makes sense (to me) to omit the first category to avoid 'double counting' and to omit the category for the reason that they are not actually part of the live site in 2026 and only still present to avoid breaking bookmarked links.

all that said, i'm happy to discuss or pivot if i'm missing anything. i expect it would be straightforward to generate a list of `/assets` urls and append them to the sitemap programmatically.

> we no longer list /links.html 🤔 🤷‍♀️

this one _certainly_ feels correct to me. i'd go as far as to say that i should open an issue to inventory more carefully and yank the page altogether in favor of a redirect if it is indeed as dead as i think it is.

curious to hear other's opinions though.

angela-tran:
@jgravois - I agree with all those reasons for omitting `/assets` and `links.html`. If we have time, it would be helpful to know what `links.html` was / is for.

I think this is ready to merge, but I'll leave it up to you if you want to check for other opinions

jgravois:
> If we have time, it would be helpful to know what `links.html` was / is for.

history seems to indicate its used in a [QR code}(https://github.com/cal-itp/calitp.org/commit/34f19bd118de35c75041fff2bb57b5ff9bb5926c). this would make a plain old redirect a perfectly sound alternative so i opened #616.

> I'll leave it up to you if you want to check for other opinions

i'll make a note of this discrepancy when we open #590 for a final round of review.

Issue #615, opened 2026-04-24
chore: configure 11ty to do the work of jekyll's seo plugin


jgravois:
## Acceptance Criteria

- [ ] the appropriate `<title>` is injected on each page
- [ ] meta tags are injected to support shareable thumbnails in slack, etc.

prior art:
* https://github.com/cal-itp/mobility-marketplace/pull/818/changes/841eb6364e7f1aba7561aeea51c35ef398f45be8

PR #617, opened 2026-04-27
fix: ensure inline links do not need to redirect

jgravois:
while working on #615 i noticed that the links in the header (and a few others) result in a 301 redirect. they're harmless, but they're still worth fixing.

Screenshot 2026-04-27 at 4 29 29 PM

### DIFF
--- a/src/_data/site.js
+++ b/src/_data/site.js
@@ -4,15 +4,15 @@ export default {
     header: [
       {
         name: "Customer stories",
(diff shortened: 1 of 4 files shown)


angela-tran (approved):
Good catch! This makes sense with how 11ty defaults to using trailing slashes  - https://www.11ty.dev/docs/permalinks/#trailing-slashes

PR #618, opened 2026-04-27
feat: add seo tags and page titles

### LINKED ISSUE #615: chore: configure 11ty to do the work of jekyll's seo plugin

jgravois:
closes #615

## Testing steps
1. compare the `<title>` and `<meta>` tags in the head of several pages w/ what is injected in https://www.calitp.org
    * homepage
    * press landing page
    * individual press page
    * /customer-stories

other than the exceptions below, the output should be the same.

## Known drift
1. `<meta property="og:description">` values now display frontmatter leads and intros
(see https://github.com/cal-itp/calitp.org/pull/618#discussion_r3150773257 for more info)
1. `datetime` values are now truncated to display only the date.
1. links pasted into slack now display a thumbnail image.

Screenshot 2026-04-28 at 12 22 18 PM

## Other fixes
1. the existing google analytics key is now injected correctly in the associated `<script>` tag.


src/_includes/meta.liquid
```diff
@@ -1,53 +0,0 @@
...
-{%- endif -%}
-
-{%- if lead -%}
-  {%- assign description = lead -%}
-{%- elsif intro -%}
-  {%- assign description = intro -%}
-{%- elsif description -%}
-  {%- assign description = description -%}
-{%- else -%}
-  {%- assign description = description -%}
```
jgravois, line 16: when i reimplemented this logic i was initially confused why i was seeing different page descriptions in this branch.

turns out i'd forgotten that [the jekyll plugin ignores the argument in favor of its own internal logic 🙃](https://github.com/cal-itp/mobility-marketplace/pull/811#discussion_r2742591536)

src/_includes/date.liquid
```diff
@@ -1 +1 @@
-<time datetime="{{ date | date_to_xmlschema }}">{{ date | date: format }}</time>
+<time datetime="{{ date |  date: "%Y-%m-%d" }}">{{ date | date: format }}</time>
```
jgravois, line 1: `date_to_xmlschema` is superfluous now because as of 6d0750e961407fc99bacf5475fc4e2227fba5a46 our front-matter dates are all already `ISO 8601` strings.

my assumption is that the `America/Los_Angeles` timezone tweak that was added in 97fbe4081c835ac457118a405c6c10c50d6ff012 was introduced to ensure that the displayed date wasn't erroneously rolled back when midnight UTC was recalculated in west coast time.

11ty has a handy mechanism of its own to account for that so i'd be up for exploring omitting timestamps altogether and just tracking plain dates (someday).

https://www.11ty.dev/docs/dates/#dates-off-by-one-day

cc/ @thekaveman

related: https://liquidjs.com/filters/date.html

src/_includes/meta.liquid
angela-tran: It took me a bit to understand what was going on here too.

The gist of it is that the `jekyll-seo-tag` plugin was using each post's `excerpt` (which by default is the post's first paragraph) as the `description` rather than this logic we had in `meta.liquid`.

See https://github.com/jekyll/jekyll-seo-tag/blob/8403350ddc7a5c7022b114ef0303d35c4dd68829/lib/jekyll-seo-tag/drop.rb#L105-L110 and https://jekyllrb.com/docs/posts/#post-excerpts.

So with the change in this PR to essentially "fix" that bug, 

https://github.com/cal-itp/calitp.org/blob/5b0e552d095ade8d169e10c0cf84285fbd1bce03/src/lib/seo.computed.js#L2

i.e. actually use the logic that reads from the page's frontmatter, **(TL;DR) the description for many pages will change to simply use the `data.site.description`.**

I think this is change is perfectly fine, just wanted to note for awareness the reason for and impact of the change.

---

**Just documenting some examples**

Here's the before and after for a page that doesn't have any of those frontmatter values - [`/fact-sheet-enhancing-transit-experience-with-bankon`](https://github.com/cal-itp/calitp.org/blob/5c92ac89d01d911aa794d273d7af9f59f8d2125b/src/resources/fact-sheet-enhancing-transit-experience-with-bankon.md):

Image

Image

<br/>

And here's an example of a page that now uses its `lead` frontmatter value instead of the previously Jekyll-generated `excerpt` - [`cal-itp-benefits-launch`](https://github.com/cal-itp/calitp.org/blob/5c92ac89d01d911aa794d273d7af9f59f8d2125b/src/press/cal-itp-benefits-launch.md):

Image

Image

angela-tran (approved):
Looks great!

PR #619, opened 2026-04-29
chore: replace links.html with simple redirect

### LINKED ISSUE #616: chore: replace /links.html with a simple redirect
we don't reference [/links.html](https://github.com/cal-itp/calitp.org/blob/f6b5b71147caaca621d438a564f79d780f6582e0/src/links.html) anywhere else in the site

because of this, it'd be cleaner to remove it entirely in favor of a simple redirect.

jgravois:
resolves #616

## Steps to test

both the urls below should resolve to the same hubspot form:

* https://www.calitp.org/links
* https://deploy-preview-619--cal-itp-website.netlify.app/links

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 # this directory has files with custom formatting we don't want prettier to revert
 docs/
 
-# this file has custom formatting we don't want prettier to revert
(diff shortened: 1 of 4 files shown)


angela-tran (approved):
🧹

PR #622, opened 2026-04-29
Chore: update Auto-PR Resources for 11ty

### LINKED ISSUE #621: Update Auto-PR creation for Resource Requests to work for 11ty
We should update the GitHub workflow / Python script that was implemented in #440 to work for the 11ty directory structure.

angela-tran:
Closes #621 

Fortunately the frontmatter for our resources stayed the same across Jekyll and 11ty, which is of note because the workflow [generates frontmatter](https://github.com/cal-itp/calitp.org/blob/f6b5b71147caaca621d438a564f79d780f6582e0/.github/scripts/resource-template.py#L182-L189) for the requested resource upload, so the only thing to update was the directory name.

### DIFF
--- a/.github/scripts/resource-template.py
+++ b/.github/scripts/resource-template.py
@@ -121,7 +121,7 @@ def build_filename(prefix, title):
   return f'{filename}.md'
 
 def create_resource(filename, content):
-  with open(f'src/_resources/{filename}', 'w') as file:
+  with open(f'src/resources/{filename}', 'w') as file:
     file.write(content)
 
 def main():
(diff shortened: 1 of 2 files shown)


jgravois (approved):
this is a great find. 🤙 

when we flip compiler.la to 11ty we'll have to remember to update the CODEOWNERS too.

https://github.com/compilerla/compiler.la/blob/dafad11ba46e5f9bb19d97bf1ccef7b24504a060/.github/CODEOWNERS#L3

angela-tran:
> this is a great find. 🤙
> 
> when we flip compiler.la to 11ty we'll have to remember to update the CODEOWNERS too.
> 
> https://github.com/compilerla/compiler.la/blob/dafad11ba46e5f9bb19d97bf1ccef7b24504a060/.github/CODEOWNERS#L3

Ah, that's a good catch too. I started a list in #623 to keep track of these post-merge items.

PR #625, opened 2026-04-30
chore: revert back to a .node-version file for maximum portability

### LINKED ISSUE #624: chore: use .node-version file to ensure dev container, pre-commit and netlify share the same runtime
when we wired up 11ty initially in https://github.com/cal-itp/calitp.org/pull/589 i fell back on using the `mise.toml` file to declare the node version because pre-commit was complaining about the .node-version setup that worked just fine for MobiMart.

it turns out this is because prettier is the only node dependent pre-commit hook and it expects/needs to see a complete version number in that file instead of just a range.

right now i'm not sure _what_ node runtime pre-commit is using, but it would be superior for the devcontainer, netlify and pre-commit to all be consistent.

jgravois:
closes #624


angela-tran (approved):

jgravois:
> **[angela-tran](https://github.com/angela-tran)** previously approved these changes on behalf of https://github.com/orgs/cal-itp/teams/web-admin [last week](https://github.com/cal-itp/calitp.org/pull/625#pullrequestreview-4221456114)

somehow i missed @angela-tran's approval last week. i _think_ this one is fine to merge into `main` so i've enabled auto-merge.

angela-tran (approved):

Issue #626, opened 2026-05-04
"Resources" in Initiatives section is missing


angela-tran:
It looks like this resources block on the homepage is missing from the 11ty version, for all the categories.

prod:
<img width="1146" height="344" alt="Screenshot 2026-05-04 at 1 30 36 PM" src="https://github.com/user-attachments/assets/1ec3986d-baf8-4239-ace7-e8900a91d037" />
<img width="1240" height="703" alt="Screenshot 2026-05-04 at 1 35 20 PM" src="https://github.com/user-attachments/assets/650edb3c-fcc7-48e2-9445-7823eaceeb4b" />

11ty:
<img width="1252" height="260" alt="Screenshot 2026-05-04 at 1 30 50 PM" src="https://github.com/user-attachments/assets/71ebfdeb-8c98-4c21-911e-7d9dc6b5b729" />
<img width="1197" height="738" alt="Screenshot 2026-05-04 at 1 34 37 PM" src="https://github.com/user-attachments/assets/6954b850-38e9-4f1f-8a4f-dba84b9a7cdc" />

_Originally posted by @cmajel in https://github.com/cal-itp/calitp.org/issues/590#issuecomment-4373118984_

PR #605, opened 2026-04-09
chore: rewire initiatives section of homepage

jgravois:
a step toward resolving #587

* https://deploy-preview-605--cal-itp-website.netlify.app/#initiatives-contactless-payments
* https://deploy-preview-605--cal-itp-website.netlify.app/#initiatives-benefits
* https://deploy-preview-605--cal-itp-website.netlify.app/#initiatives-gtfs

PR #627, opened 2026-05-04
Fix: Initiatives Resources section

angela-tran:
Closes #626 

The Initiatives Resources section needed the same refactor as what was done in #605 for the Initiatives Press section.


jgravois (approved):
nice catch @cmajel. ⚾ 

i don't know how i missed this the first time around, i probably mistakenly assumed that "See all" was supposed to be the only link shown for resources. 🤦‍♂️ 

thx for the cleanup @angela-tran! 😇

Issue #629, opened 2026-05-04
fix: link directly to external content when applicable in press landing page


jgravois:
> When clicking through on press pages that go to external sites, I noticed the 11ty version would redirect slower, and and shows a flash of a "redirecting" page, but this does not happen on prod (from what I can tell). Is that intended? 

_Originally posted by @cmajel in https://github.com/cal-itp/calitp.org/issues/590#issuecomment-4373257556_

looks like currently in the list of press content we're linking to our own internal url. if the item is ultimately an external link, that's the url that we should use in the `<a>` tag.

<img width="1121" height="565" alt="Image" src="https://github.com/user-attachments/assets/3867eb3a-ea7b-4398-994d-81df722f61bf" />

## ie:
/press/sacrts-new-tap2-ride-contactless-payment-system-and-veterans-discount-fare-pilot-program-launching-soon/

should be:
https://www.sacrt.com/sacrts-new-tap2ride-contactless-payment-system-and-veterans-discount-fare-pilot-program-launching-soon/

jgravois:
fixed in #631

PR #631, opened 2026-05-04
fix: reference 'external' front-matter for press items correctly

jgravois:
closes #629

### DIFF
--- a/src/_includes/article.liquid
+++ b/src/_includes/article.liquid
@@ -9,7 +9,7 @@
   {% endif %}
 {% else %}
   {% assign url = item.url %}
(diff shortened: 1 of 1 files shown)


src/_includes/article.liquid
```diff
@@ -9,7 +9,7 @@
   {% endif %}
 {% else %}
   {% assign url = item.url %}
```
jgravois, line 11: the reason `item.url` resolves is because its a piece of metadata that 11ty itself hydrates for us.

angela-tran (approved):
🎉

PR #638, opened 2026-05-11
Fix: use null alt text for decorative images

### LINKED ISSUE #628: Remove alt text for decorative images
There are a number of images on the home page that have alt text that describe the images as decorative. This can be removed for a `null` alt text value per [WCAG guidelines](https://www.w3.org/WAI/tutorials/images/decorative/).

Example:
https://github.com/cal-itp/calitp.org/pull/590/changes/BASE..2bb6f89bbde96f2731231cf8f3017ffb1c10187d#diff-f5f97dee597184c9e3f55f21916a8c2b8c149762bd85e52149a31b77cf5cb59bR33

angela-tran:
Closes #628

### DIFF
--- a/src/index.liquid
+++ b/src/index.liquid
@@ -30,14 +30,14 @@ stylesheet: homepage.css
   <img
     class="w-100"
     src="images/home-divider-1.png"
-    alt="Decorative element with dots and dashes, meant to resemble a transit map"
+    alt=""
   >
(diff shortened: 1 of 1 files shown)


jgravois (approved):
TIL that `alt=""` is more helpful than omitting the attribute entirely.

Issue #613, opened 2026-04-23
DSDL: sketch out a simple versioning system


jgravois:
right now MobiMart loads whatever copy of DSDL it finds on the `main` branch of this repository.

https://github.com/cal-itp/mobility-marketplace/blob/da5b5a1090b70079d924b0003ffcaa99c785f78c/src/_layouts/main.html#L21

as part of expanding its scope and implementing DSDL in Benefits we can/should rely on git tags for this job. that will allow us to reference an [immutable copy](https://raw.githubusercontent.com/cal-itp/calitp.org/refs/tags/2023.08.1/src/stylesheets/main.css) of DSDL from consumer sites and free ourselves up to make breaking changes on `main`.

as an extra credit task, we might as well add a simple publish github actions workflow that publishes a new release when a calendar version is detected.

prior art:
https://github.com/cal-itp/benefits/blob/c19939679e108a3c446042e3cb10f5638b27bfef/.github/workflows/deploy.yml#L113-L131

## Acceptance criteria

- [ ] an initial tag is pushed that contains DSDL to date
  makes sense (to me) to do this as soon as #590 is merged
- [ ] open MobiMart PR to load DSDL from that tag instead of from `main`
- [ ] non pre-release tags trigger a github release via a new github action
- [ ] https://docs.calitp.org/calitp.org/ includes reference documentation related to versioning
- [ ] calitp.org itself continues to load DSDL from `main` acting as a canary for new/experimental changes

(if we were starting from scratch, i'd reach for SemVer, but given that [we already have one calendar release](https://github.com/cal-itp/calitp.org/releases/tag/2023.08.1) in this repository, i don't see a need to switch now)

angela-tran:
Here's a recap of what @jgravois and I talked through earlier today to plan out a versioning system for DSDL.

## First, a description of where we currently are:

In #640, a GitHub Actions workflow was introduced to create a new [release](https://github.com/cal-itp/calitp.org/releases) of calitp.org whenever we push a CalVer-formatted tag of this repo. 

Note that when the release is created doesn't necessarily correspond with when new changes are deployed to the live calitp.org -- that happens on any merge to `main`. New changes could be live on the site without there being a GitHub release that documents those changes right away. 
So you could think of a release as more of a "formal" announcement of changes that have gone out on calitp.org as well as serving as the version marker for that snapshot of DSDL.

[...]

PR #640, opened 2026-05-12
chore: add docs and github action for creating calendar versions

jgravois:
this PR tackles the first two TODO:s in #613.

view it live: https://calitp-website-640--cal-itp-previews.netlify.app/guides/dsdl/release/

### DIFF
--- a/docs/guides/dsdl/release.md
+++ b/docs/guides/dsdl/release.md
@@ -1,0 +1,23 @@
+# DSDL releases
+
+## Publishing a release
+
+The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
+
+## Consuming a release
+
+Referencing a specific [release](https://github.com/cal-itp/calitp.org/releases) via a link tag is recommended:
+
+```html
+<link
+  rel="stylesheet"
+  href="https://raw.githubusercontent.com/cal-itp/calitp.org/refs/tags/2026.03.1/src/stylesheets/dsdl/dsdl.css"
+>
+```
+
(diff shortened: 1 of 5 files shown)


angela-tran (approved):
Sorry this sat around for a bit. I had skimmed it before and probably should've just approved it then

Docs look good!

docs/guides/dsdl/release.md
```diff
@@ -0,0 +1,23 @@
...
+The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
+
+## Consuming a release
+
+Referencing a specific [release](https://github.com/cal-itp/calitp.org/releases) via a link tag is recommended:
+
+```html
+<link
+  rel="stylesheet"
+  href="https://raw.githubusercontent.com/cal-itp/calitp.org/refs/tags/2026.03.1/src/stylesheets/dsdl/dsdl.css"
```
angela-tran, line 14: I'm curious if it's ok to assume that this service provides reliable hosting / that it's ok to hotlink from `raw.githubusercontent.com`

angela-tran: I think we should forge ahead with the path you're laying down, and we can deal with issues if they come up, so don't mind me. 😅 But yeah, I think this was what was giving me hesitation.

jgravois: i wouldn't call it 'bulletproof', but my understanding is that it is just as dependable as the gh-pages/github hosting we're relying on now.

https://github.com/cal-itp/mobility-marketplace/blob/11bca90da7b8784deb8b54156af396ac8a78cb40/src/_layouts/main.html#L21

related: https://github.com/rgrove/rawgit/issues/197

jgravois: if raw.githubusercontent.com ever lets us down, something like https://www.jsdelivr.com/?docs=gh would be worth investigating.

its possible that's just a pass-through CDN anyway though.

PR #642, opened 2026-05-13
squash of branch head for h2h comparison

FreedomFaighter:
fix: scrub mention of defunct mailing list

gitignore update, bin moved to scripts, pre-commit reference to bin changed to scripts


thekaveman:
Thank you, we already updated the defunct newsletter links in #641.

PR #650, opened 2026-05-28
Chore: add instructions for expected handling of asset files

### LINKED ISSUE #649: Update Resource Auto-PR workflow to match latest procedure
@jgravois I'm thinking we should create a ticket to update the workflow that creates these PRs so that it:
- doesn't add the Google Drive link (and maybe leaves a note to upload the PDF)

## Additional context
`resource-template.yml` is the workflow

_Originally posted by @angela-tran in https://github.com/cal-itp/calitp.org/pull/648#issuecomment-4565186780_

angela-tran:
Closes #649 

In #566, there was some clarification needed on the latest guidance for how the asset file should be processed. Guidance was documented in https://github.com/cal-itp/calitp.org/pull/566#issuecomment-3921776652 and in the README via #575, but on the auto-created PRs themselves, it's not clear that this is the expectation.

This PR adds some steps to the generated PR description to help make it more obvious and reduce doubt / confusion.

### DIFF
--- a/.github/workflows/resource-template.yml
+++ b/.github/workflows/resource-template.yml
@@ -59,7 +59,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           title: "Auto-PR for Resource Request #${{ steps.workflow_state.outputs.issue_number }}"
           draft: true
-          body: >
+          body: |-
             This PR is created automatically from a resource request. Please
             review the changes and merge when ready.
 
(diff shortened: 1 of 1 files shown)


jgravois (dismissed):
i'm lovin it. 🍔 

if we want to preserve the whitespace above and below your new block i think we need to switch from the `>` to the `|-` block scalar, but i wouldn't call that a blocker. 😇 

https://yaml-multiline.info/

angela-tran:
Great suggestion @jgravois, thanks!

jgravois (approved):
🌈🌈

PR #652, opened 2026-06-03
chore: constrain devcontainer user permissions

### LINKED ISSUE #645: chore: update devcontainer setup to use non-root user
lets emulate the [benefits devcontainer](https://github.com/cal-itp/benefits/blob/baea4e0018802b53a04254a0130bf965bca5c5ca/.devcontainer/Dockerfile#L2) configuration and ensure that root privileges are inaccessible after the container is built.

since we're not pulling docker-python-web, we'll need to do the same legwork as is done here 👇

https://github.com/compilerla/compiler.la/blob/main/.devcontainer/Dockerfile#L7-L8

jgravois:
closes #645

## How i tested

running `su root` now returns a prompt to supply a password. 🔒 

confirmed that npm scripts can still be executed, new JS deps can still be installed and pre-commit hooks are still functioning.

## Potpourri

while i was in there i tacked on two extra commits to 🛁 
* resolve one `npm audit` nag
* bump to Node.js 24.16.0 (the latest stable minor release)
* tweak the post install script to keep the npm cli up to date instead of relying on what comes bundled with node
* ensure our local spellcheck pre-commit step introspects changed files in addition to the commit message


angela-tran (approved):
This looks good and works for me 👍 

I ran into a bunch of permission errors when trying to build this branch using my existing copy of this repo and then remembered that Docker handles ownership differently across [Mac, Windows, and Linux](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user) when it mounts folders/files.

Image


The easiest way for me to get rid of my lingering `root`-owned folders/files was to just do a fresh clone of this repo, immediately switch to this branch, and then build to get a successful devcontainer build. ✅

angela-tran:
Also:
- saw updated Node version (24.16.0) :white_check_mark: 
- saw that cspell checks spelling of changed file contents :white_check_mark:

PR #656, opened 2026-06-08
Chore: use same Liquid extension as other static site projects

### LINKED ISSUE #654: Normalize formatting config and Liquid extension to that of other static site repos
After looking over all 3 of the static site repos whose auto-formatting we've been poking at over the past few months (calitp.org, MobiMart, and compiler.la), I've identified these 2 items as the last little bit of clean-up. Luckily it was only this repo that had lingering clean-up to do.

- [x] In `.vscode/settings.json`, fix the overrides for `editor.defaultFormatters` to not use multi-language syntax - #655 
  - Just in case you missed it, these overrides are to account for developers having their own language-specific user settings, perhaps unknowingly. Ideally, we could just assume no one has any surreptitiously-added user settings and simply rely on `"editor.defaultFormatter": "esbenp.prettier-vscode",`, but we've been bitten by this [in the wild](https://github.com/compilerla/compiler.la/pull/300#issuecomment-4345000354) more than once, so we decided to just be paranoid like this.
  - This resolves https://github.com/cal-itp/mobility-marketplace/pull/912#discussion_r3320657271 
    > (as a follow-up, we should fix that calitp.org settings.json file to separate out the languages
  - https://github.com/compilerla/compiler.la/pull/300#issuecomment-4343932761 is where we discovered "VS Code does not support [combined language syntax](https://code.visualstudio.com/docs/configure/settings#_multiple-languagespecific-editor-settings) for `editor.defaultFormatter`"
- [x] In `.devcontainer/devcontainer.json`, use the `shopify.theme-check-vscode` extension instead of `sissel.shopify-liquid` - #656 
  - This is what all our other static site repos are using. The origin of this switch was PR [cal-itp/mobility-marketplace#824](https://github.com/cal-itp/mobility-marketplace/pull/824/changes#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R26) back in February

angela-tran:
Closes #654 

Addresses the second item listed in the ticket.

MobiMart: 
https://github.com/cal-itp/mobility-marketplace/blob/cb4603856a7c575599407a8e5c98ac8396de3225/.devcontainer/devcontainer.json#L25

compiler.la: 
https://github.com/compilerla/compiler.la/blob/e717571f5b8263019926430d6f68d73a189add7a/.devcontainer/devcontainer.json#L25

(side note: why doesn't GitHub do the cool snippet preview thing anymore when you paste permalinks to lines of code? :frowning_face:)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,7 +23,7 @@
         "esbenp.prettier-vscode",
         "mhutchie.git-graph",
         "redhat.vscode-xml",
-        "sissel.shopify-liquid",
+        "shopify.theme-check-vscode",
         "streetsidesoftware.code-spell-checker",
         "tombi-toml.tombi"
       ]


Scotchester (approved):

Scotchester:
> (side note: why doesn't GitHub do the cool snippet preview thing anymore when you paste permalinks to lines of code? ☹️)

Hmm, that is odd… 🤔

jgravois:
> why doesn't GitHub do the cool snippet preview thing

i've only ever seen snippet previews for code in the same repository.

PR #663, opened 2026-06-15
chore: install node in Dockerfile for better caching

### LINKED ISSUE #661: chore: ensure node installation survives a devcontainer rebuild
the way things are set up right now Node.js isn't installed until the `postAttach.sh` script is run.

it would be more performant to ensure this happens within the Dockerfile itself so that the step is cached for subsequent builds.

jgravois:
closes #661

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -6,7 +6,7 @@ RUN useradd --create-home --shell /bin/bash $USER
 ENV WORK_DIR=/home/$USER/site
 WORKDIR $WORK_DIR
 
-RUN chown -R $USER $WORK_DIR
+RUN chown -R $USER /home/$USER
 
 # Install mise (to manage Node.js) https://mise.jdx.dev/mise-cookbook/docker.html#docker-image-with-mise
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -33,9 +33,6 @@ USER $USER
 RUN git config --global --add safe.directory $WORK_DIR
 
 # Install Node.js and npm dependencies
-COPY mise.toml mise.toml
-COPY .node-version .node-version
-COPY package.json package.json
-COPY package-lock.json package-lock.json
+COPY mise.toml .node-version package.json package-lock.json ./
 RUN mise exec -- npm install -g npm@latest
 RUN mise exec -- npm clean-install
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,11 +1,12 @@
 FROM python:3.12
 
 ENV USER=calitp
+RUN useradd --create-home --shell /bin/bash $USER
 
-RUN useradd --create-home --shell /bin/bash $USER && \
-    chown -R $USER /home/$USER
+ENV WORK_DIR=/home/$USER/site
+WORKDIR $WORK_DIR
 
-WORKDIR /home/$USER/site
+RUN chown -R $USER $WORK_DIR
 
 # Install mise (to manage Node.js) https://mise.jdx.dev/mise-cookbook/docker.html#docker-image-with-mise
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
@@ -13,6 +14,7 @@ ENV MISE_DATA_DIR="/home/$USER/mise"
 ENV MISE_CONFIG_DIR="/home/$USER/mise"
 ENV MISE_CACHE_DIR="/home/$USER/mise/cache"
 ENV MISE_INSTALL_PATH="/usr/local/bin/mise"
+ENV MISE_TRUSTED_CONFIG_PATHS=$WORK_DIR
 ENV PATH="/home/$USER/mise/shims:$PATH"
 
 RUN curl https://mise.run | sh
@@ -27,3 +29,13 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
 RUN pip install --no-cache-dir -r docs/requirements.txt
 
 USER $USER
+
+RUN git config --global --add safe.directory $WORK_DIR
+
+# Install Node.js and npm dependencies
+COPY mise.toml mise.toml
+COPY .node-version .node-version
+COPY package.json package.json
+COPY package-lock.json package-lock.json
+RUN mise exec -- npm install -g npm@latest
+RUN mise exec -- npm clean-install
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -19,8 +19,6 @@ ENV PATH="/home/$USER/mise/shims:$PATH"
 
 RUN curl https://mise.run | sh
 
-RUN apt-get update
-
 # Install Python dependencies
 COPY .devcontainer/requirements.txt .devcontainer/requirements.txt
 COPY docs/requirements.txt docs/requirements.txt
@@ -34,5 +32,4 @@ RUN git config --global --add safe.directory $WORK_DIR
 
 # Install Node.js and npm dependencies
 COPY mise.toml .node-version package.json package-lock.json ./
-RUN mise exec -- npm install -g npm@latest
-RUN mise exec -- npm clean-install
+RUN mise exec -- npm install -g npm@latest && npm clean-install
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,9 +4,6 @@
   "service": "site",
   "runServices": ["site", "docs"],
   "workspaceFolder": "/home/calitp/site",
-  "remoteEnv": {
-    "MISE_TRUSTED_CONFIG_PATHS": "${containerEnv:PATH}:/home/calitp/site"
-  },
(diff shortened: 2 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -4,9 +4,6 @@
   "service": "site",
   "runServices": ["site", "docs"],
   "workspaceFolder": "/home/calitp/site",
-  "remoteEnv": {
-    "MISE_TRUSTED_CONFIG_PATHS": "${containerEnv:PATH}:/home/calitp/site"
-  },
```
jgravois, line 9: i don't know _why_, but when i moved more logic into the Dockerfile itself, `/home/calitp/site` stopped being trusted until i started setting the same environment variable there explicitly.
```dockerfile
ENV MISE_TRUSTED_CONFIG_PATHS=$WORK_DIR
```
it feels like an improvement to have all the setup happen in the same file, but i would have expected this to work too 🤷

.devcontainer/Dockerfile
```diff
@@ -1,18 +1,20 @@
...
 ENV USER=calitp
+RUN useradd --create-home --shell /bin/bash $USER
 
-RUN useradd --create-home --shell /bin/bash $USER && \
-    chown -R $USER /home/$USER
+ENV WORK_DIR=/home/$USER/site
+WORKDIR $WORK_DIR
 
-WORKDIR /home/$USER/site
+RUN chown -R $USER $WORK_DIR
```
jgravois: same here. in my own testing, `npm install` failed to write files to disk because of a lack of permissions until i updated the `chown` call to reference the `/site` directory specifically and waited until after the directory had been created.

.devcontainer/Dockerfile
```diff
@@ -27,3 +29,13 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
 RUN pip install --no-cache-dir -r docs/requirements.txt
 
 USER $USER
+
+RUN git config --global --add safe.directory $WORK_DIR
```
jgravois, line 31: this is an unrelated change, but we keep finding that recent versions of `git` are fussier about trust and consolidating these calls in the Dockerfile helps us sidestep hardcoding the same work directory in different files.

see https://github.com/cal-itp/mobility-marketplace/pull/924#discussion_r3358541469

.devcontainer/devcontainer.json
Scotchester: I think it might have worked if you used `containerEnv`? See the definitions in the `devcontainer.json` spec: https://containers.dev/implementors/json_reference/#general-properties

.devcontainer/Dockerfile
Scotchester: Did you try `chown`ing just `/home/$USER`, but after creating the `site` dir?

.devcontainer/devcontainer.json
jgravois: ah, that makes sense! makes me wonder whether the way we were specifying `${containerEnv:PATH}` in `remoteEnv` was a dead-end. 🤔 

either way, i think what i'm suggesting here is an improvement. happy to discuss further if anyone else disagrees though.

Scotchester: Yeah, I think I prefer it in the `Dockerfile` anyway :)

.devcontainer/Dockerfile
```diff
@@ -27,3 +29,13 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
...
+
+# Install Node.js and npm dependencies
+COPY mise.toml mise.toml
+COPY .node-version .node-version
+COPY package.json package.json
+COPY package-lock.json package-lock.json
```
Scotchester: This could be simplified, if you like:

```suggestion
COPY mise.toml .node-version package.json package-lock.json ./
```

When more than two arguments are given, all but the last argument are source files, and the last argument is the destination folder. [[ref](https://docs.docker.com/reference/dockerfile/#source-1)]

.devcontainer/Dockerfile
jgravois: when i did, i must have had something else wrong 😇 

7c25674

what you're suggesting is the right move because there are things going on _outside_ of `/site`, but i still don't understand why existing call isn't sufficient.

Scotchester: Re: previous behavior – I am not a Unix permissions expert, but my assumption is that regardless of the `calitp` user owning `/home/calitp`, a new subdirectory being created by `root` will be owned by `root`.

.devcontainer/Dockerfile
```diff
@@ -27,3 +29,13 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
...
+RUN mise exec -- npm install -g npm@latest
+RUN mise exec -- npm clean-install
```
Scotchester: I don't have a theory as to why yet, but `node_modules` is not being saved in the built image. We don't have a `.dockerignore` file, and as far as I can tell, there's no such thing as a global or default `.dockerignore` it might be looking at, so I don't know why they wouldn't be getting in there.

jgravois: huh. i could see the directory locally but after blowing it away manually and rebuilding (w/o cache) i can see that you're right.

great catch. i'll have to look into this one.

jgravois: looks like this'll fix it. i'm going to do a little more research before i commit the change though.

```yml
services:
  site:
    volumes:
      - ..:/home/calitp/site
      - /home/calitp/site/node_modules
```

jgravois: done in 3cd758a

Scotchester: Huh. Do you know why that would fix it?

Scotchester (dismissed):
I don't know why, but the last commit DOES fix builds from scratch with no local `node_modules`. Seems like this is good to go.

.devcontainer/Dockerfile
jgravois: I am far from an expert, so take this explanation with a grain of salt.

my understanding is that the build phase filesystem is entirely isolated; not just from the parent machine but also the built output image. because of this we have to explicitly create a volume if we'd like to expose the contents of `node_modules` downstream.

ref: https://stackoverflow.com/a/61810781

I tried to summarize this in the code comment, but feedback is (always) welcome

Scotchester: Thanks, I saw the comment after I asked the question, but it still didn't totally click so I let the question stand. I still don't feel like I understand, but I don't think I need to since it works :)

.devcontainer/Dockerfile
lalver1: Maybe we can remove this line since it doesn't look like we are using `apt`?

.devcontainer/Dockerfile
```diff
@@ -27,3 +29,10 @@ RUN pip install --no-cache-dir -r .devcontainer/requirements.txt
 RUN pip install --no-cache-dir -r docs/requirements.txt
 
 USER $USER
+
+RUN git config --global --add safe.directory $WORK_DIR
+
+# Install Node.js and npm dependencies
+COPY mise.toml .node-version package.json package-lock.json ./
+RUN mise exec -- npm install -g npm@latest
```
lalver1: We can probably consolidate the 2 `RUN` commands to something like

```
RUN mise exec -- npm install -g npm@latest && \
    mise exec -- npm clean-install
```

to make the image smaller (and maybe faster to build?).

lalver1 (commented):
Looks good @jgravois! I have a few comments in case you think it's worth to make these changes. The image build time goes down by 4 seconds and the image is 200MB smaller, so not a huge difference but maybe worth it.

I also want to quickly look at Docker's built in caching feature for `pip` and `npm` and see if it's worth implementing here. We use it for Benefits so it might be good to use it here too.

.devcontainer/Dockerfile
jgravois: as long as i stared at this Dockerfile this never even occurred to me 😆

.devcontainer/Dockerfile
jgravois: i liked this idea so much i turned it into a one-liner. c6199eb

i'm only seeing the overall image size shrink from 2.35 GB to 2.31 GB with the changes you're suggesting, but i think they're good tidying regardless.

are you clocking image build time with a stopwatch or is there a podman or docker utility for that that i'm unaware of?

lalver1: Nice! Yep, not huge differences but more like tidying up/consistency across our Dockerfiles.

So I'm building the dev container's image using something like

```
time docker build --no-cache -t calitp-site:cache_3 -f .devcontainer/Dockerfile .
```

from the project's root. It gives you timing results like this

```
real    0m0.221s
user    0m0.075s
sys     0m0.056s
```

and this [thread](https://stackoverflow.com/questions/556405/what-do-real-user-and-sys-mean-in-the-output-of-time1) has a quick explanation of what they mean. I think you use zsh, but I think the `time` command is also available there.

I went looking at Benefits and how we use [Docker's cache for `pip`](https://github.com/cal-itp/benefits/blob/main/.devcontainer/Dockerfile#L39) and we could do something similar here (also for `npm`) but maybe that's out of scope for this PR and won't get us a whole lot of savings.

lalver1 (approved):
Looks great @jgravois! Built it locally and served the site without a problem 👍

PR #664, opened 2026-06-15
docs: mention netlify permalinks in release publishing guide

### LINKED ISSUE #613: DSDL: sketch out a simple versioning system

### LINKED ISSUE #662: Update DSDL release reference doc to reflect new versioning process
https://docs.calitp.org/calitp.org/guides/dsdl/release/ should include a table that maps DSDL versions to their Netlify permalinks. We also need a note on how for now, whoever releases a new DSDL version should find the Netlify permalink and add it to the table.

jgravois:
closes #662

### DIFF
--- a/docs/guides/dsdl/release.md
+++ b/docs/guides/dsdl/release.md
@@ -4,14 +4,17 @@
 
 The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
 
+After a release has been published it is necessary to identify the associated netlify deployment and use the permalink to update the list below.
+
 ## Consuming a release
 
 Referencing a specific [release](https://github.com/cal-itp/calitp.org/releases) via a link tag is recommended:
 
+### `2026.05.01`
 ```html
 <link
   rel="stylesheet"
-  href="https://raw.githubusercontent.com/cal-itp/calitp.org/refs/tags/2026.03.1/src/stylesheets/dsdl/dsdl.css"
+  href="https://6a0e339e78e8ea00086cacb6--cal-itp-website.netlify.app/stylesheets/dsdl/dsdl.css"
 >
 ```
 
--- a/docs/guides/dsdl/release.md
+++ b/docs/guides/dsdl/release.md
@@ -4,7 +4,7 @@
 
 The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
 
-After a release has been published it is necessary to identify the associated netlify deployment and use the permalink to update the list below.
+After a release has been published, it is necessary to identify the associated Netlify deployment and document the release's permalink on this page.
 
 ## Consuming a release
 


docs/guides/dsdl/release.md
```diff
@@ -4,14 +4,17 @@
 
 The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
 
+After a release has been published it is necessary to identify the associated netlify deployment and use the permalink to update the list below.
```
angela-tran: @jgravois  - re: https://github.com/cal-itp/calitp.org/issues/662#issuecomment-4712842834,

> when i sat down and looked at this it felt like a more helpful v1 would be to just let folks copy/paste the entire link tag out of the documentation instead of displaying one generic snippet and then appending a lookup table and asking folks to inject the url itself from there.

Yeah, I can see how it kinda feels premature to write this page like we're gonna have a bunch of versions.

Maybe the snippet could always have the latest URL? And the table or list is where the older versions would go.

I'm ok with not having the table or list right now, and if that's the case, could we re-word the last part of this sentence? Maybe something like:

```suggestion
After a release has been published, it is necessary to identify the associated Netlify deployment and document the release's permalink on this page.
```

jgravois: these are all great suggestions 🤙

angela-tran (approved):
📚

jgravois:
this work was ultimately unwound. see https://github.com/cal-itp/mobility-marketplace/pull/944 for more info.

PR #668, opened 2026-06-25
docs: clarify that it is not yet possible to reference tagged DSDL

jgravois:
this PR more or less reverts #664

### DIFF
--- a/docs/guides/dsdl/release.md
+++ b/docs/guides/dsdl/release.md
@@ -4,23 +4,10 @@
 
 The steps to create a release tag are identical to [Benefits](https://docs.calitp.org/benefits/guides/release/). Release candidate tags are skipped.
 
-After a release has been published, it is necessary to identify the associated Netlify deployment and document the release's permalink on this page.
+## Consuming DSDL
 
-## Consuming a release
-
-Referencing a specific [release](https://github.com/cal-itp/calitp.org/releases) via a link tag is recommended:
-
-### `2026.05.01`
-```html
-<link
-  rel="stylesheet"
-  href="https://6a0e339e78e8ea00086cacb6--cal-itp-website.netlify.app/stylesheets/dsdl/dsdl.css"
->
-```
-
-Loading bleeding-edge CSS (directly from [`main`](https://github.com/cal-itp/calitp.org/blob/main/src/stylesheets/dsdl/dsdl.css)) is _not_ recommended.
+A mechanism to consume previously tagged DSDL releases does not exist yet. For now the easiest option is to load the bleeding-edge CSS directly from [`main`](https://github.com/cal-itp/calitp.org/blob/main/src/stylesheets/dsdl/dsdl.css).
 
 ```html
-<!-- don't do this -->
 <link rel="stylesheet" href="https://www.calitp.org/stylesheets/dsdl/dsdl.css">
 ```


docs/guides/dsdl/release.md
```diff
@@ -4,23 +4,10 @@
...
-After a release has been published, it is necessary to identify the associated Netlify deployment and document the release's permalink on this page.
+## Consuming DSDL
 
-## Consuming a release
-
-Referencing a specific [release](https://github.com/cal-itp/calitp.org/releases) via a link tag is recommended:
-
-### `2026.05.01`
-```html
-<link
```
thekaveman, line 15: I still think having the example HTML snippet, with the link to the live CSS file, could be useful here.

docs/guides/dsdl/release.md
```diff
@@ -4,23 +4,10 @@
...
-  rel="stylesheet"
-  href="https://6a0e339e78e8ea00086cacb6--cal-itp-website.netlify.app/stylesheets/dsdl/dsdl.css"
->
-```
-
-Loading bleeding-edge CSS (directly from [`main`](https://github.com/cal-itp/calitp.org/blob/main/src/stylesheets/dsdl/dsdl.css)) is _not_ recommended.
+A mechanism to consume previously tagged DSDL releases does not exist yet. For now the easiest option is to load the bleeding-edge CSS directly from [`main`](https://github.com/cal-itp/calitp.org/blob/main/src/stylesheets/dsdl/dsdl.css).
```
angela-tran, line 9: This `github.com` URL takes us to the GitHub UI showing the file rather than the actual CSS. And since we can't use the raw GitHub content URL (`https://raw.githubusercontent.com/cal-itp/calitp.org/refs/heads/main/src/stylesheets/dsdl/dsdl.css`), I think our only choice is to use the bleeding-edge CSS from calitp.org?

```suggestion
A mechanism to consume previously tagged DSDL releases does not exist yet. For now the easiest option is to load the bleeding-edge CSS directly from [`main`](https://www.calitp.org/stylesheets/dsdl/dsdl.css).
```

angela-tran: Or maybe you meant to link to the file rather than provide a link to the raw CSS :sweat_smile:

docs/guides/dsdl/release.md
jgravois: i definitely left that in place, because i agree. 😇 

https://calitp-website-668--cal-itp-previews.netlify.app/guides/dsdl/release/

docs/guides/dsdl/release.md
jgravois: the code snippet directly below is a valid technique to load the actual CSS from a downstream app.

i thought it'd be helpful to link to github to show the same file in the context of version control, but i don't feel strongly about it.

docs/guides/dsdl/release.md
thekaveman: 🤦 my bad, didn't scroll down far enough in the diff

docs/guides/dsdl/release.md
angela-tran: Sounds good, I think I just read this too quickly. Linking to the file in GitHub makes sense :+1:

angela-tran (approved):

